PR #13, opened 2023-09-26
Make CPU architecture options shorter

MaxDesiatov:
`--host-cpu-architecture` and `--target-cpu-architecture` are quite verbose options, `--target-arch` and `--host-arch` seem to be nicer to use as shorthands.

### DIFF
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -60,7 +60,6 @@ struct GeneratorCLI: AsyncParsableCommand {
   var linuxDistributionVersion: String?
 
   @Option(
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

kateinoigakukun:
I feel they are too short, also `-h` looks like something to print help info How about `--target-arch` or `--target-cpu`? Clang and Rust compiler use target-cpu to select a specific CPU architecture.

MaxDesiatov:
I prefer `--target-arch` to `--target-cpu`, since the latter feels like specifying a concrete CPU model like `M1`, instead of triple architectures like `arm64.

kateinoigakukun (approved):
Seems better to me :)

PR #14, opened 2023-09-26
Update to the latest tag of AsyncAlgorithms

MaxDesiatov:
There have been no releases of AsyncAlgorithms for some time until now with `1.0.0-alpha`, which apparently provides a bugfix for `combineLatest` operator that we use. It doesn't look like `from: "1.0.0"` allows SwiftPM to pick up this prerelease version, so I've specified it as `exact`.


MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):

PR #16, opened 2023-09-26
Allow downloading LLVM project source for building `lld`

MaxDesiatov:
First step towards fixing https://github.com/apple/swift-sdk-generator/issues/9. To build `lld` on `x86_64` we need to download its sources first, and smallest source distributions are available at https://github.com/llvm/llvm-project/releases.

`DownloadableArtifacts` was also refactored to group hashes for different artifacts into fewer dictionaries.

### DIFF
--- a/Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
+++ b/Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
@@ -13,9 +13,55 @@
 import struct Foundation.URL
 import struct SystemPackage.FilePath
 
-/// SHA256 hashes of Swift packages for Linux known to the generator.
-private let knownLinuxSwiftVersions: [LinuxDistribution: [String: [Triple.CPU: String]]] = [
-  LinuxDistribution.ubuntu(.jammy): [
+enum ArtifactOS: Hashable {
+  init(_ tripleOS: Triple.OS, _ versions: VersionsConfiguration) {
+    switch tripleOS {
+    case .linux:
+      self = .linux(versions.linuxDistribution)
+    case .macosx, .darwin:
+      self = .macOS
+    }
+  }
+
+  case linux(LinuxDistribution)
+  case macOS
+  case source
+
+  var llvmBinaryURLSuffix: String {
+    switch self {
+    case .linux: "linux-gnu"
+    case .macOS: "apple-darwin22.0"
+    case .source: fatalError()
+    }
+  }
+}
+
+typealias CPUMapping = [Triple.CPU: String]
+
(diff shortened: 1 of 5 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/Generator/LocalSwiftSDKGenerator.swift
```diff
@@ -20,7 +20,7 @@ public final class LocalSwiftSDKGenerator: SwiftSDKGenerator {
   public let artifactID: String
   public let versionsConfiguration: VersionsConfiguration
   public let pathsConfiguration: PathsConfiguration
-  public let downloadableArtifacts: DownloadableArtifacts
+  public var downloadableArtifacts: DownloadableArtifacts
```
MaxDesiatov, line 23: Marked mutable as `useLLVMSources` is mutating.

Sources/SwiftSDKGenerator/Generator/LocalSwiftSDKGenerator.swift
```diff
@@ -40,6 +40,7 @@ public final class LocalSwiftSDKGenerator: SwiftSDKGenerator {
       .removingLastComponent()
       .removingLastComponent()
       .removingLastComponent()
+      .removingLastComponent()
```
MaxDesiatov, line 43: Fix for `Bundles` ending up in the `Sources` directory after `LocalSwiftSDKGenerator.swift` was moved to a subdirectory.

MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
```diff
@@ -13,71 +13,88 @@
...
+    case .macosx, .darwin:
+      self = .macOS
+    }
+  }
+
+  case linux(LinuxDistribution)
+  case macOS
+
+  /// No specific OS required for this artifact, it's not binary and is distributed in the source form.
+  case source
```
kateinoigakukun, line 31: Is this used anywhere? If it's not used now and we don't have a corresponding return value on this case for `llvmBinaryURLSuffix`, would it make sense to remove this case?

kateinoigakukun (approved):

Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
MaxDesiatov: Good catch, I've cleaned that up.

MaxDesiatov:
@swift-ci test

PR #17, opened 2023-09-26
Fix formatting, rename `Destination.swift`

MaxDesiatov:
In accepted [SE-0387](https://github.com/apple/swift-evolution/blob/main/proposals/0387-cross-compilation-destinations.md) "destinations" were renamed to "Swift SDKs". Based on that it feels more suitable to rename `Destination.swift` to `SwiftSDKMetadata.swift`. Also taking this as an opportunity to apply formatting cleanups and to update `Utilities/soundness.sh` to exclude the generated `./Bundles` directory from scanning.


MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):

PR #19, opened 2023-09-28
Build `lld` when no prebuilt binaries are available

MaxDesiatov:
Resolves https://github.com/apple/swift-sdk-generator/issues/9.
Resolves rdar://115890915.

### DIFF
--- a/Sources/SwiftSDKGenerator/Generator/LocalSwiftSDKGenerator.swift
+++ b/Sources/SwiftSDKGenerator/Generator/LocalSwiftSDKGenerator.swift
@@ -344,6 +344,21 @@ public final class LocalSwiftSDKGenerator: SwiftSDKGenerator {
     }
   }
 
+  public func buildCMakeProject(_ projectPath: FilePath) async throws -> FilePath {
+    try await Shell.run(
+      """
+      PATH='/bin:/usr/bin:\(Self.homebrewPrefix)/bin' \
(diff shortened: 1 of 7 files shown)


MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):

Sources/SwiftSDKGenerator/Generator/LocalSwiftSDKGenerator.swift
```diff
@@ -344,6 +344,21 @@ public final class LocalSwiftSDKGenerator: SwiftSDKGenerator {
     }
   }
 
+  public func buildCMakeProject(_ projectPath: FilePath) async throws -> FilePath {
+    try await Shell.run(
```
kateinoigakukun, line 348: It's fine to me to use Shell based command invocation for now, but do you have any plan to switch to Process based invocation to avoid unintentional shell parameter expansion?

MaxDesiatov: Yes, I would love to, unfortunately `ProcessExecutor` from the `AsyncProcess` module here doesn't support the `currentWorkingDirectory` API.

MaxDesiatov: We would need to evaluate our use of `currentWorkingDirectory` first, and check if there's a way to avoid any uses of it. Should be possible for both `cmake` and `swift build` invocations, but there are a few more commands remaining that I haven't checked yet.

MaxDesiatov:
@swift-ci test

PR #22, opened 2023-10-02
Fix `.createNew` deprecation in `EventLoopGroupProvider`

MaxDesiatov:
Latest release of AHC (0.19.0) deprecated `EventLoopGroupProvider.createNew`, we should be using `.singleton` instead.


MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):
Looks reasonable to me :+1:

PR #25, opened 2023-10-10
Use linuxDistributionVersion if specified otherwise use default version for linuxDistributionName

rimim:
Command line argument --linux-distribution-version was ignored and a default version was always used for the specified --linux-distribution-name.


MaxDesiatov:
@swift-ci test

PR #28, opened 2023-10-25
Remove protocol indirection for `SwiftSDKGenerator`

MaxDesiatov:
There is only a single implementation this protocol: `LocalSwiftSDKGenerator`. The original intent was to introduce additional implementations for testing, but I no longer think that this protocol on its own is useful. We should protocolize on a different level, i.e. file system access, process spawning etc. That will be realized in subsequent PRs.

This also allowed removing `public` access modifiers from a lot of properties and functions that didn't actually need to be exposed outside of the `SwiftSDKGenerator` module.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):
I totally agree with this direction

MaxDesiatov:
@swift-ci test

PR #29, opened 2023-10-25
Add `SystemSQLite` target for future use in caching

MaxDesiatov:
In subsequent PRs I'd like to introduce a SQLite-backed cache for storing hashes of generated artifacts. That will help us in avoiding redundant work and will make generator significantly quicker for re-runs that change only a few arguments.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -60,6 +60,7 @@ let package = Package(
         .enableExperimentalFeature("StrictConcurrency=complete"),
       ]
     ),
+    .systemLibrary(name: "SystemSQLite", pkgConfig: "sqlite3"),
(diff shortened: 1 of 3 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

kateinoigakukun:
Can we vendor amalgamation sqlite.c instead of depending on system library?

MaxDesiatov:
I need to clarify that. But in the meantime we may need to update Linux CI Dockerfiles to preinstall it.

kateinoigakukun (approved):

PR #30, opened 2023-10-25
[NFC] Use `https://` instead of `http://` in license headers

MaxDesiatov:
IIRC certain browsers even mark http:// URLs as not secure by default.


MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):

PR #31, opened 2023-10-25
Add `GeneratorEngine` module with tests and macros

MaxDesiatov:
We'd like to avoid redundant rebuilding and regeneration for inputs that don't change. The new `GeneratorEngine` module achieves that by introducing a `Query` abstraction, which can be hashed and matched to its cached outputs if any are available.

The new `Engine` actor can run such `Query` and immediately return a cached result file if one is present and its hash matches the recorded one.

This required introducing new `@Query` and `@CacheKey` macros for making hashing easy and consistent on `QueryProtocol` types and their inputs.  We can't use the standard `Hashable` protocol, as that produces new hashes on every process launch due to randomized hashing seeds. With Swift Crypto's `HashFunction` we can generate consistent hashes and record them in `SQLiteBackedCache`.

To make the caching engine testable, I've introduced a new `FileSystem` actor protocol with `VirtualFileSystem` and `LocalFileSystem` actor implementations.

### DIFF
--- a/Sources/GeneratorEngine/Cache/CacheKeyProtocol.swift
+++ b/Sources/GeneratorEngine/Cache/CacheKeyProtocol.swift
@@ -24,12 +24,14 @@ public protocol CacheKeyProtocol {
 
 extension Bool: CacheKeyProtocol {
   public func hash(with hashFunction: inout some HashFunction) {
+    "Swift.Bool".hash(with: &hashFunction)
     hashFunction.update(data: self ? [1] : [0])
   }
 }
 
 extension Int: CacheKeyProtocol {
   public func hash(with hashFunction: inout some HashFunction) {
+    "Swift.Int".hash(with: &hashFunction)
     withUnsafeBytes(of: self) {
       hashFunction.update(bufferPointer: $0)
     }
@@ -38,6 +40,10 @@ extension Int: CacheKeyProtocol {
 
 extension String: CacheKeyProtocol {
   public func hash(with hashFunction: inout some HashFunction) {
+    var t = "Swift.String"
+    t.withUTF8 {
+      hashFunction.update(bufferPointer: .init($0))
+    }
     var x = self
     x.withUTF8 {
       hashFunction.update(bufferPointer: .init($0))
@@ -47,20 +53,26 @@ extension String: CacheKeyProtocol {
 
 extension FilePath: CacheKeyProtocol {
   public func hash(with hashFunction: inout some HashFunction) {
-    self.description.hash(with: &hashFunction)
+    "SystemPackage.FilePath".hash(with: &hashFunction)
+    self.string.hash(with: &hashFunction)
   }
 }
 
 extension URL: CacheKeyProtocol {
   public func hash(with hashFunction: inout some HashFunction) {
+    "Foundation.URL".hash(with: &hashFunction)
     self.description.hash(with: &hashFunction)
   }
 }
 
 extension Optional: CacheKeyProtocol where Wrapped: CacheKeyProtocol {
   public func hash(with hashFunction: inout some HashFunction) {
+    "Swift.Optional".hash(with: &hashFunction)
     if let self {
+      true.hash(with: &hashFunction)
       self.hash(with: &hashFunction)
+    } else {
+      false.hash(with: &hashFunction)
     }
   }
 }
--- a/Sources/GeneratorEngine/Cache/CacheKeyProtocol.swift
+++ b/Sources/GeneratorEngine/Cache/CacheKeyProtocol.swift
@@ -1,0 +1,69 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Macros
+
+@_exported import protocol Crypto.HashFunction
+import struct Foundation.URL
+import struct SystemPackage.FilePath
+
+/// Indicates that values of a conforming type can be hashed with an arbitrary hashing function. Unlike `Hashable`,
+/// this protocol doesn't utilize random seed values and produces consistent hash values across process launches.
+public protocol CacheKeyProtocol {
+  func hash(with hashFunction: inout some HashFunction)
+}
+
+extension Bool: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    hashFunction.update(data: self ? [1] : [0])
+  }
+}
+
+extension Int: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    withUnsafeBytes(of: self) {
+      hashFunction.update(bufferPointer: $0)
+    }
+  }
+}
+
+extension String: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    var x = self
+    x.withUTF8 {
+      hashFunction.update(bufferPointer: .init($0))
+    }
+  }
+}
+
+extension FilePath: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    self.description.hash(with: &hashFunction)
+  }
+}
+
+extension URL: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    self.description.hash(with: &hashFunction)
+  }
+}
+
+extension Optional: CacheKeyProtocol where Wrapped: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    if let self {
+      self.hash(with: &hashFunction)
+    }
+  }
+}
+
+@attached(extension, conformances: CacheKeyProtocol, names: named(hash(with:)))
+public macro CacheKey() = #externalMacro(module: "Macros", type: "CacheKeyMacro")
--- a/Sources/GeneratorEngine/Cache/SQLite.swift
+++ b/Sources/GeneratorEngine/Cache/SQLite.swift
@@ -1,0 +1,310 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2014-2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import struct Foundation.Data
+import SystemPackage
+import SystemSQLite
+
+extension FilePath: @unchecked Sendable {}
+
+/// A minimal SQLite wrapper.
+public final class SQLite {
+  enum Error: Swift.Error, Equatable {
+    case databaseFull
+    case message(String)
+  }
+
+  /// The location of the database.
+  let location: Location
+
+  /// The configuration for the database.
+  let configuration: Configuration
+
+  /// Pointer to the database.
+  let db: OpaquePointer
+
+  /// Create or open the database at the given path.
+  ///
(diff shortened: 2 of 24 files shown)


MaxDesiatov:
@swift-ci test

Sources/GeneratorEngine/Cache/SQLite.swift
MaxDesiatov, line 1: Vendored from TSC with minimal changes.

Sources/GeneratorEngine/Cache/SQLiteBackedCache.swift
MaxDesiatov, line 1: Vendored from TSC with minimal changes.

Sources/GeneratorEngine/FileSystem/FileLock.swift
MaxDesiatov, line 1: Vendored from TSC with minimal changes.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
Curious to hear @weissi's opinion on this too. This has no proper CAS yet, just hashing of files produced by queries, but I'm making changes here as incremental as possible. We can get a local CAS-like storage eventually with more generator code rewritten with a query-based approach.

Sources/GeneratorEngine/Cache/CacheKeyProtocol.swift
```diff
@@ -0,0 +1,69 @@
...
+    var x = self
+    x.withUTF8 {
+      hashFunction.update(bufferPointer: .init($0))
+    }
+  }
+}
+
+extension FilePath: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    self.description.hash(with: &hashFunction)
```
weissi, line 50:
```suggestion
    self.string.hash(with: &hashFunction)
```

Sources/GeneratorEngine/Cache/CacheKeyProtocol.swift
```diff
@@ -0,0 +1,69 @@
...
+
+extension URL: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    self.description.hash(with: &hashFunction)
+  }
+}
+
+extension Optional: CacheKeyProtocol where Wrapped: CacheKeyProtocol {
+  public func hash(with hashFunction: inout some HashFunction) {
+    if let self {
```
weissi, line 62: that's not ideal. You want to do hash

`hashOf(0 or false)` if `nil`. And `hashOf(1 or true) ++ hashOf(self)` if not nil.

Otherwise you'll get hash unnecesary hash collisions

Sources/GeneratorEngine/Cache/SQLite.swift
weissi: I'd recommend to use Vapor's sqlite-nio over this

Sources/GeneratorEngine/Engine.swift
```diff
@@ -0,0 +1,81 @@
...
+    cacheLocation: SQLite.Location
+  ) {
+    self.fileSystem = fileSystem
+    self.httpClient = httpClient
+    self.logger = logger
+    self.resultsCache = SQLiteBackedCache(tableName: "cache_table", location: cacheLocation, logger)
+  }
+
+  deinit {
+    try! resultsCache.close()
```
weissi, line 47: this is fraught with peril. Not guaranteed to happen, might block etc. Would very much recommend to make the lifecycle explicit and to assert in deinit that it has been shut down.

Sources/GeneratorEngine/Engine.swift
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  /// Executes a given query if no cached result of it is available. Otherwise fetches the result from engine's cache.
+  /// - Parameter query: A query value to execute.
+  /// - Returns: A file path to query's result recorded in a file.
+  public subscript(_ query: some QueryProtocol) -> FilePath {
+    get async throws {
+      var hashFunction = SHA512()
+      query.hash(with: &hashFunction)
+      let key = hashFunction.finalize().description
```
weissi, line 57: `.description`'s output is usually not API. Is this truly the right thing to do?

Sources/GeneratorEngine/FileSystem/FileLock.swift
weissi: do you really need file locks? Those have very bad and annoying properties

Sources/GeneratorEngine/FileSystem/LocalFileSystem.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SystemPackage
+
+public actor LocalFileSystem: FileSystem {
+  public static let defaultChunkSize = 128 * 1024
```
weissi, line 16: I'd recommend to up this to at least 512kiB

Sources/GeneratorEngine/FileSystem/ReadableFileStream.swift
```diff
@@ -0,0 +1,106 @@
...
+
+      let bytesRead = try buffer.withUnsafeMutableBytes {
+        try self.fileDescriptor.read(into: $0)
+      }
+
+      guard bytesRead > 0 else {
+        return nil
+      }
+
+      return .init(buffer[0..<bytesRead])
```
weissi, line 71: this allocates and copies

weissi: you probably want to removeLast(n) instead

Sources/GeneratorEngine/FileSystem/ReadableFileStream.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    deinit {
+      try! fileDescriptor.close()
```
weissi, line 75: this is a very bad idea. Managing file descriptors with `deinit` is all but guaranteed to break everything.

Make the lifecycle explicit or you'll suffer hard. Also `close` should be `async`.

I'd recommend `try await withOpenFile(...) { file in try await file.write(...) }` such that you can't forget the closes.

Tests/GeneratorEngineTests/EngineTests.swift
```diff
@@ -0,0 +1,152 @@
...
+import XCTest
+
+private let encoder = JSONEncoder()
+private let decoder = JSONDecoder()
+
+private extension FileSystem {
+  func read<V: Decodable>(_ path: FilePath, as: V.Type) async throws -> V {
+    let fileStream = try await self.read(path)
+    var bytes = [UInt8]()
+    for try await chunk in fileStream {
```
weissi, line 26: I'd ask the user to provide a limit here and maybe default it to 10 MiB or something. Limitless memory ballooning is never a good API.

Tests/GeneratorEngineTests/EngineTests.swift
```diff
@@ -0,0 +1,152 @@
...
+      bytes += chunk
+    }
+
+    return try decoder.decode(V.self, from: .init(bytes))
```
weissi, line 30: - this allocates and copies from `Array` -> `Data`. Would recommend to accumulate in `Data`
- would recommend to never use `.init`, it obfuscates the code a lot. Some `.init`s allocate & copy, others don't, very very hard to see here what's going on.

Tests/GeneratorEngineTests/EngineTests.swift
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  func write(_ path: FilePath, _ value: some Encodable) async throws {
+    let data = try encoder.encode(value)
+    try await self.write(path, .init(data))
```
weissi, line 35: allocate & copy here too (might be hard to avoid)

MaxDesiatov:
macOS CI currently blocked by https://github.com/apple/swift/issues/69485, so we'll have to either force merge or disable the macOS check temporarily

Sources/GeneratorEngine/Cache/SQLite.swift
MaxDesiatov: I'd like this code to be easily portable to Windows, and IIUC sqlite-nio will make that much harder for me.

Sources/GeneratorEngine/FileSystem/FileLock.swift
MaxDesiatov: What would be a good alternative for that?

MaxDesiatov:
@swift-ci test

Sources/GeneratorEngine/Engine.swift
MaxDesiatov: I've changed `SQLiteBackedCache` type to take `some Sequence<UInt8>` as keys to improved the situation. But there are still types like `FileCacheRecord` that are encoded into JSON and we need to serialize hashes as strings for that. But that stems from the fact that `SQLiteBackedCache` is a NoSQL store at this stage. I'll normalize it to make it properly relational in a future PR.

MaxDesiatov:
@swift-ci test

Sources/GeneratorEngine/Cache/CacheKeyProtocol.swift
euanh: @MaxDesiatov You don't seem to have addressed this comment.   All the others are now showing as 'Outdated'.   Are you planning to make any more changes?

euanh: @MaxDesiatov Does the code here address @weissi 's comment?

MaxDesiatov: Not sure why it doesn't show "outdated" here, I've updated it this way to address the comment:
```swift
extension Optional: CacheKeyProtocol where Wrapped: CacheKeyProtocol {
  public func hash(with hashFunction: inout some HashFunction) {
    "Swift.Optional".hash(with: &hashFunction)
    if let self {
      true.hash(with: &hashFunction)
      self.hash(with: &hashFunction)
    } else {
      false.hash(with: &hashFunction)
    }
  }
}
```

euanh: I thought that might be it, but I was having trouble following the PR history to check.

MaxDesiatov:
@swift-ci test

euanh (approved):
I've built and tested an SDK with this changeset.   The non-Docker case works, and not rebuilding `lld` is a welcome speedup.

I'm having problems with `--with-docker` SDKs, both for RHEL and Ubuntu:

```
% swift build --experimental-swift-sdk 5.9-RELEASE_ubuntu_jammy_x86_64
Building for debugging...
warning: Could not read SDKSettings.json for SDK at: /Users/euanh/Library/org.swift.swiftpm/swift-sdks/5.9-RELEASE_ubuntu_jammy_x86_64.artifactbundle/5.9-RELEASE_ubuntu_jammy_x86_64/x86_64-unknown-linux-gnu/ubuntu-jammy.sdk
<unknown>:0: warning: glibc not found for 'x86_64-unknown-linux-gnu'; C stdlib may be unavailable
<unknown>:0: warning: glibc not found for 'x86_64-unknown-linux-gnu'; C stdlib may be unavailable
warning: Could not read SDKSettings.json for SDK at: /Users/euanh/Library/org.swift.swiftpm/swift-sdks/5.9-RELEASE_ubuntu_jammy_x86_64.artifactbundle/5.9-RELEASE_ubuntu_jammy_x86_64/x86_64-unknown-linux-gnu/ubuntu-jammy.sdk
error: link command failed with exit code 1 (use -v to see invocation)
ld.lld: error: cannot open Scrt1.o: No such file or directory
ld.lld: error: cannot open crti.o: No such file or directory
ld.lld: error: unable to find library -lc
ld.lld: error: cannot open crtn.o: No such file or directory
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
error: fatalError
[5/6] Linking hello-world
```
However I can reproduce these problems on `main`, even going back to https://github.com/apple/swift-sdk-generator/commit/321e22a46db6b5744284c40a5aac758d7e0b9ab2.  There might be something wrong with my local setup, but it certainly doesn't seem to be a regression caused by this change.   I'll continue investigating.

PR #34, opened 2023-10-30
Cache `DownloadableArtifacts` with `GeneratorEngine`

MaxDesiatov:
New `GeneratorEngine` allows us to cache all items in `DownloadableArtifacts` in a generic way and to allow redundant re-downloads for artifacts that don't have their checksums hardcoded (e.g. LLVM sources for `lld`).

This now allows us to delete all hardcoded checksums and also to cache other artifacts in the future, like `lld` binary and even whole artifact bundles.


Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
```diff
@@ -45,74 +46,26 @@ enum ArtifactOS: Hashable {
...
     let isPrebuilt: Bool
   }
 
   let hostSwift: Item
   private(set) var hostLLVM: Item
   let targetSwift: Item
 
-  let allItems: [Item]
+  private let shouldUseDocker: Bool
+  var allItems: [Item] {
```
MaxDesiatov, line 62: Fixing a bug here, where `allItems` wasn't recomputed after running `useLLVMSources`.

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -98,7 +98,13 @@ struct GeneratorCLI: AsyncParsableCommand {
...
         isVerbose: self.verbose
       )
-      .generateBundle(shouldGenerateFromScratch: !self.incremental)
+      do {
+        try await generator.generateBundle(shouldGenerateFromScratch: !self.incremental)
+        try await generator.shutDown()
+      } catch {
+        try await generator.shutDown()
+        throw error
+      }
```
euanh, line 107: It would be nice add a `withGenerator` wrapper which could take care of shutting down and rethrowing if necessary.

Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
```diff
@@ -45,74 +46,26 @@ enum ArtifactOS: Hashable {
 
 typealias CPUMapping = [Triple.CPU: String]
 
-/// SHA256 hashes of binary LLVM artifacts known to the generator.
-private let knownLLVMBinariesVersions: [ArtifactOS: [String: CPUMapping]] = [
-  .macOS: [
-    "15.0.7": [
-      Triple.CPU.arm64: "867c6afd41158c132ef05a8f1ddaecf476a26b91c85def8e124414f9a9ba188d",
```
euanh, line 52: A side effect of using these hashes was that we also verified that the upstream tarballs hadn't changed from the known versions.   LLVM signs its binaries using GPG, so in future we could have an option to verify those signatures.   This would require the user to download and validate LLVM's public key.

MaxDesiatov: I agree, checking signatures is the right way to do it. But code on `main` that handled checksums was inconsistent and didn't warn users on checksums mismatch. It would only redownload new archives from remote URLs on a mismatch or if a checksum was absent. From that perspective, we didn't regress, and the caching behaviour is actually improved.

euanh (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

PR #35, opened 2023-10-31
Cache `lld` binaries to avoid redundant rebuilds

MaxDesiatov:
Depends on https://github.com/apple/swift-sdk-generator/pull/34.

This proceeds with converting more of the generation steps to a query-based architecture. Now `lld` doesn't need to be rebuilt if its source directory and build options don't change. This is implemented with a new `CMakeBuildQuery`, which takes source directory, and build options as arguments, and returns a freshly-built binary that can be cached by the generator engine.

Also renamed `DownloadQuery` to `DownloadArtifactQuery` and moved it to a separate file.


euanh (approved):

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import GeneratorEngine
+import struct SystemPackage.FilePath
+
+@Query
+struct CMakeBuildQuery {
+  let sourcesDirectory: FilePath
+  let outputBinarySubpath: [FilePath.Component]
```
euanh, line 19:
```suggestion
  let outputBinarySubpath: FilePath
```

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
```diff
@@ -0,0 +1,35 @@
...
+      """
+      cmake -B build -G Ninja -S llvm -DCMAKE_BUILD_TYPE=Release \(self.options)
+      """,
+      currentDirectory: self.sourcesDirectory
+    )
+
+    let buildDirectory = self.sourcesDirectory.appending("build")
+    try await Shell.run("ninja", currentDirectory: buildDirectory)
+
+    return self.outputBinarySubpath.reduce(into: buildDirectory) { $0.append($1) }
```
euanh, line 33:
```suggestion
    return buildDirectory.appending(outputBinarySubpath.compontents)
```

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
MaxDesiatov: Not sure how suitable `FilePath` is for storing relative paths, `[FilePath.Component]` seemed to be a better options for this.

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
MaxDesiatov: This was mostly about making it explicit that the query takes relative paths as arguments and not absolute ones.

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Unpack.swift
```diff
@@ -66,36 +68,43 @@ extension SwiftSDKGenerator {
...
-        pathsConfiguration.toolchainBinDirPath.appending("wasm-ld")
-      default:
-        fatalError()
-      }
+    let unpackedLLDPath = if llvmArtifact.isPrebuilt {
+      untarDestination.appending("bin/lld")
+    } else {
+      try await self.engine[CMakeBuildQuery(
+        sourcesDirectory: untarDestination,
+        outputBinarySubpath: ["bin", "lld"],
```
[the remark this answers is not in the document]
MaxDesiatov: The danger here is that an absolute path can be supplied in the string literal and we won't be able to catch it at compile time.

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
euanh: Yes, I was reading the docs for FilePath just now and I wondered if that was what you were intending to do, to make sure components such as `bin/lld` aren't mixed up with full paths.

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
euanh: Yes, I was reading the docs for FilePath just now and I wondered if that was what you were intending to do, to make sure components such as `bin/lld` aren't mixed up with full paths.

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
MaxDesiatov, line 19:
```suggestion
  /// Path to the output binary relative to the CMake build directory.
  let outputBinarySubpath: [FilePath.Component]
```

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
MaxDesiatov: I've added a doc comment on that `CMakeBuildQuery` property explicitly mentioning relative paths, which I hope will clarify things a bit.

Sources/SwiftSDKGenerator/Queries/CMakeBuildQuery.swift
```diff
@@ -0,0 +1,36 @@
...
```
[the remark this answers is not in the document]
MaxDesiatov: Great, thanks, I'll update this to become more readable! I'd be happy to merge after that, but prefer to merge PRs separately, so I need a review on the engine PR at the start of this PR chain: https://github.com/apple/swift-sdk-generator/pull/31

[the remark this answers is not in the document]
euanh: I see.   I wondered if you were going to merge them all back down into `max/engine` and keep that branch going for a while, but I see you have now addressed Johannes's comments.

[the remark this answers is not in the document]
MaxDesiatov: Yeah, I prefer merging PRs from a stack separately, especially as I stick to "Squash and merge" that allows to preserve history of each PR in separate commits.

MaxDesiatov:
@swift-ci test

PR #36, opened 2023-11-01
Fix cross-compilation with static linking when using C interop

MaxDesiatov:
Cross-compiling with Swift SDKs produced with Swift SDK Generator with `--static-swift-stdlib` passed to `swift build` in projects that use C interop (e.g. via `import Glibc`) leads to these errors:

```
<module-includes>:1:10: note: in file included from <module-includes>:1:
#include "SwiftGlibc.h"
```

The reason was that underlying Clang headers were present in `usr/lib/swift/clang` subdirectory of the `swift.xctoolchain` directory of the final bundle, but `usr/lib/swift_static/clang` was missing. These directories are identical, we should symlink one to the other for static linking to work.

Resolves rdar://115784447.


MaxDesiatov:
@swift-ci test

tomerd (approved):

PR #37, opened 2023-11-01
Remove unused Darwin platforms from the host toolchain

MaxDesiatov:
We've removed these subdirectories specified in `unusedDarwinPlatforms` constant previously from `usr/lib/swift`, but kept in `usr/lib/swift_static` by accident. Cleaning them up should reduce the size of the final Swift SDK bundle slightly.


MaxDesiatov:
@swift-ci test

tomerd (approved):

MaxDesiatov:
@swift-ci test

PR #38, opened 2023-11-01
Cache downloads of `.deb` packages w/ `DownloadFileQuery`

MaxDesiatov:
This shaves off ~30s from the Ubuntu target generator run time on my machine, bringing it down to ~1:30 with all currently available caching applied. We can go faster than this, but that will require support for hashing directory contents or multiple files in the engine in a future PR, as the remaining potentially queryfiable tasks produce directories or multiple files as outputs.


MaxDesiatov:
@swift-ci test

PR #46, opened 2023-11-09
Update `AsyncProcess`, use it in `Shell` implementation

MaxDesiatov:
These changes make `AsyncProcess` implementation more robust and increase its code coverage.

I had to update Swift NIO dependencies, since updates to `AsyncProcess` relies on some new NIO APIs.

Now we no longer have to rely on `Foundation.Process` directly in `Shell`, as `AsyncProcess` abstracts rough edges away for us and also provides an async-friendly API.

Prerequisite of rdar://116204896.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/SystemUtils/Shell.swift
```diff
@@ -130,61 +106,40 @@ final class Shell {
...
+    if shouldLogCommands {
+      print(command)
+    }
+
+    let result = try await ProcessExecutor.runCollectingOutput(
+      executable: "/bin/sh",
+      ["-c", command],
+      collectStandardOutput: true,
+      collectStandardError: false,
+      perStreamCollectionLimitBytes: 100 * 1024 * 1024
```
euanh, line 122:
```suggestion
      perStreamCollectionLimitBytes: 100 * 1024 * 1024,
      environment: ProcessInfo.processInfo.environment
```
We need to inherit the environment here, otherwise the `sh` process will run with a default path of `"/usr/gnu/bin:/usr/local/bin:/bin:/usr/bin:."`.   This is sufficient for many of the commands we call using `readStdout`, such as `xar`, `rsync` and even `swift`, but it will cause `launchDockerCommand` to fail because `docker` is in the Homebrew path.

(The other functions using `docker`, such as `copyFromDockerContainer`, don't have this problem because they use `Shell.run`, which _does_ inherit the environment.)

euanh (changes requested):

Sources/SwiftSDKGenerator/SystemUtils/Shell.swift
MaxDesiatov: Nice catch, thanks!

MaxDesiatov:
@swift-ci test

euanh (commented):
The `AsyncProcess`-based version of `Shell.run` handles output differently to the old version on `main`, so we seem to have lost a couple of pieces of log output when compared to the `main` branch:
* `docker cp` used to print a one-line live output counting up how many megabytes had been copied from the container.   Now there is no output.
* The `lld` build used to print a one-line live output counting each edge as it was built.  Now each line is printed individually (~1500 lines of output, ignoring the double spacing)

I played with setting `standardError`/`standardOutput` to `.inherit` but it wasn't immediately successful.   A bit more digging may be necessary if we want to retain the live output.

The changes seem just to be cosmetic - the resulting SDKs are ok.

Sources/SwiftSDKGenerator/SystemUtils/Shell.swift
```diff
@@ -130,61 +106,41 @@ final class Shell {
...
-          pipe.fileHandleForReading.readabilityHandler = nil
-        }
-      }
-
-      process[keyPath: pipeKeyPath] = pipe
+extension ChunkSequence {
+  func printChunksAsStrings() async throws {
+    for try await var chunk in self {
+      guard let string = chunk.readString(length: chunk.readableBytes) else { continue }
+      print(string)
```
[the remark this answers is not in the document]
weissi: Couple of things:
- no need to `var chunk`, please `let chunk`
- instead of `guard let string = chunk.readString(length: chunk.readableBytes) else { continue }` please use `String(buffer: chunk)`, easier and better
- if you want to print line by line it'd be

```
for try await let line in self.splitIntoLines() {
    print(line)
}
```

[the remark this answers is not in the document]
weissi: That should correct the output and is also much shorter.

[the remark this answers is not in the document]
weissi: Alternatively, there are a bunch of helpers such as

```
        try await ProcessExecutor.runProcessingOutput(
            executable: "/bin/foo",
            [ "bar", "buz" ],
            outputProcessor: { stream, line in
                print(line, "(from \(stream))")
            },
            splitOutputIntoLines: true,
        ).throwIfNonZero()
```

the `outputProcessor` allows you to specify a quick closure if your processing is very limited

euanh (approved):

Sources/SwiftSDKGenerator/SystemUtils/Shell.swift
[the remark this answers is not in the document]
weissi: Or of course

```swift
            let exe = ProcessExecutor(
                executable: "/usr/bin/foo", ["bar"],
            )
            async let result = exe.run()
            for try await line in await merge(exe.standardOutput.splitIntoLines(dropTerminator: true).strings, exe.standardError.splitIntoLine(dropTerminator: true).strings) {
                print(line)
            }
            try await result.throwIfNonZero() // throw if process exited non-zero
```

if you don't want to use the `runProcessingOutput` convenience

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
```diff
@@ -147,17 +147,12 @@ extension HTTPClient {
...
       throw FileOperationError.downloadFailed(URL(string: url)!)
     }
 
-    var result = ""
-    for try await chunk in try packages.unzip(isVerbose: isVerbose) {
-      result.append(String(data: chunk, encoding: .utf8)!)
-    }
-
-    return result
+    return packages.readString(length: packages.readableBytes)
```
weissi, line 155: please just `return String(buffer: packages)` and make it a `let packages`

Sources/SwiftSDKGenerator/SystemUtils/ByteBuffer+Utils.swift
```diff
@@ -10,16 +10,23 @@
...
-    let gzip = try Shell("gzip -cd", shouldLogCommands: isVerbose)
-    gzip.stdin.write(Data(buffer: self))
-    try gzip.stdin.close()
+  func unzip(isVerbose: Bool) async throws -> ByteBuffer? {
+    let result = try await ProcessExecutor.runCollectingOutput(
+      executable: "/usr/bin/gzip", ["-cd"],
+      standardInput: [self].async,
+      collectStandardOutput: true,
+      collectStandardError: false,
+      perStreamCollectionLimitBytes: 100 * 1024 * 1024
```
weissi, line 25: that seems a bit much, that's 100 MiB, no?

Sources/SwiftSDKGenerator/SystemUtils/GeneratorError.swift
```diff
@@ -15,19 +15,29 @@ import enum NIOHTTP1.HTTPResponseStatus
 import struct SystemPackage.FilePath
 
 enum GeneratorError: Error {
+  case noProcessOutput(String)
+  case unhandledChildProcessSignal(Int32, CommandInfo)
```
weissi, line 19:
```suggestion
  case unhandledChildProcessSignal(CInt, CommandInfo)
```

Sources/SwiftSDKGenerator/SystemUtils/GeneratorError.swift
```diff
@@ -15,19 +15,29 @@ import enum NIOHTTP1.HTTPResponseStatus
...
+  case nonZeroExitCode(Int32, CommandInfo)
```
weissi, line 20:
```suggestion
  case nonZeroExitCode(CInt, CommandInfo)
```

Sources/SwiftSDKGenerator/SystemUtils/Shell.swift
```diff
@@ -130,61 +106,41 @@ final class Shell {
...
+    if shouldLogCommands {
+      print(command)
+    }
+
+    let result = try await ProcessExecutor.runCollectingOutput(
+      executable: "/bin/sh",
+      ["-c", command],
+      collectStandardOutput: true,
+      collectStandardError: false,
+      perStreamCollectionLimitBytes: 100 * 1024 * 1024,
```
weissi, line 122: 100 MiB?

Sources/SwiftSDKGenerator/SystemUtils/Shell.swift
```diff
@@ -130,61 +106,41 @@ final class Shell {
...
 
-    try await process.waitUntilExit()
+    try result.exitReason.throwIfNonZero()
 
-    var output = ""
-    for try await chunk in process.stdout {
-      output.append(String(data: chunk, encoding: .utf8)!)
+    guard
+      var buffer = result.standardOutput,
+      let result = buffer.readString(length: buffer.readableBytes)
```
weissi, line 130: `let result = String(buffer: buffer)`, no need for the `guard`

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
weissi: see the docs: https://swiftinit.org/docs/swift-nio/niocore/bytebuffer#Basic%20usage

wholesale _conversion_ of ByteBuffer is always `TargetType(buffer: myByteBuffer)` Like `Array(buffer: byteBuffer)`, `String(buffer: myByteBuffer)`, ...

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/SystemUtils/ByteBuffer+Utils.swift
MaxDesiatov: Would 10 MiB be a better choice? It's also not entirely clear if that's the top limit for collected bytes altogether or only for the intermediate buffer? What happens when the collected amount exceeds the limit?

MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/SystemUtils/ByteBuffer+Utils.swift
weissi: It's the per stream collection limit. As in: How many bytes will we collect for stdout / stderr (they're both allowed to reach the limit) before stopping and producing an error.

10 MiB isn't necessarily better but 100 MiB is quite a high limit. Usually if you need significantly more than 1 MiB or you should be streaming the data rather than ballooning it into memory.

PR #47, opened 2023-11-10
Cancel tasks when receiving a `SIGINT` signal

MaxDesiatov:
This adds a dependency on the `ServiceLifecly` product from https://github.com/swift-server/swift-service-lifecycle.

Had to add a manual call to `withTaskCancellationHandler` in the current `Shell` implementation to propagate `SIGINT` to child processes on task cancellation.

I've also made `Engine` initializer and `shutDown` method internal, using `withEngine` pattern instead, which fixes shut down bugs that came up with task cancellation.

Also fixed `ArchitectureMappingTests.swift` not building due to recent changes and macOS CI being broken, while tests are temporarily disabled on Linux CI (tracked in https://github.com/apple/swift-sdk-generator/issues/45, reenabled in a subsequent PR).

Resolves rdar://116204896.

### DIFF
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -14,7 +14,6 @@ import ArgumentParser
 import Logging
 import ServiceLifecycle
 import SwiftSDKGenerator
-import UnixSignals
 
 @main
 struct GeneratorCLI: AsyncParsableCommand {
@@ -91,13 +90,7 @@ struct GeneratorCLI: AsyncParsableCommand {
   )
   var targetArch: Triple.CPU? = nil
 
-  enum State {
-    case generatorFinished
-    case generatorFailed(Error)
-    case signalReceived(UnixSignal)
-  }
-
-  mutating func run() async throws {
+  func run() async throws {
     let linuxDistributionDefaultVersion = switch self.linuxDistributionName {
     case .rhel:
       "ubi9"
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -11,7 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
+import AsyncAlgorithms
 import SwiftSDKGenerator
+import UnixSignals
 
 @main
 struct GeneratorCLI: AsyncParsableCommand {
@@ -88,6 +90,12 @@ struct GeneratorCLI: AsyncParsableCommand {
   )
   var targetArch: Triple.CPU? = nil
 
+  enum State {
+    case generatorFinished
+    case generatorFailed(Error)
+    case signalReceived(UnixSignal)
+  }
+
   mutating func run() async throws {
     let linuxDistributionDefaultVersion = switch self.linuxDistributionName {
     case .rhel:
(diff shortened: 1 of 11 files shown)


MaxDesiatov:
@swift-ci test

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -11,7 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 import ArgumentParser
+import AsyncAlgorithms
```
MaxDesiatov, line 14:
```suggestion
```

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

FranzBusch:
> The reason for keeping these tasks unstructured is that we need to hold handles to these for explicit cancellation. Then if a signal is produced by `UnixSignalSequence` first, we can cancel the generator task. If the generator finishes successfully, we cancel the signals reader task and shut down normally. If the generator fails with an error, we rethrow it from the task to the top level.

This can be achieved by using `Service` and `ServiceGroup` from `ServiceLifecycle`. Just make your sub systems such as the generator task conform to `Service` and provide a `run` method. Afterwards you can put them into the `ServiceGroup` and map the signals you care about to graceful shutdown. Then inside your `Service` you just have to implement graceful shutdown probably for you that is just cancelling so you can wrap your code in `cancelOnGracefulShutdown {}`. Importantly, graceful shutdown already does what you manually do here which is in-reverse order shutdown of the individual services. Only once a service has exited cleanly the `ServiceGroup` is going to shutdown the next one. Lastly, you can customise the termination behaviour of individual services if you want to. This allows you to configure what happens if a service exits or throws unexpectedly.

MaxDesiatov:
@swift-ci test

euanh:
Before this change, if I pressed `ctrl-c` while ninja was building `lld`, the swift-sdk-generator task would be killed but ninja would keep running and would have to be killed separately.
With this change, `ctrl-c` doesn't kill `swift-sdk-generator` while ninja is running - ctrl-c is ignored.  Is that expected?  I was expecting that ctrl-c would kill swift-sdk-generator _and_ ninja with this change.

MaxDesiatov:
Sorry, I should've marked it as a draft, there's `withTaskCancellationHandler` missing here that I'm adding and testing right now.

MaxDesiatov:
> With this change, `ctrl-c` doesn't kill `swift-sdk-generator` while ninja is running - ctrl-c is ignored. Is that expected?

That's fixed now, and child processes are also sent `SIGINT`, ready for review.

MaxDesiatov:
@swift-ci test

euanh:
`Ctrl-C` now stops all the child processes.

If you cancel during the download phase the generator stops but prints at least some of the log messages for the unpacking step.   Similar things happen if you cancel during the later steps.  I'm not sure whether or not the `xar` task below was started or this was just a log message, but these messages are a little confusing.

If the user cancels we should probably not print the `Time taken for this generator run: 2 seconds.` message, or change it to `This generator run cancelled after: 2 seconds`.

```
Using downloaded artifacts in these locations:
/Users/euanh/sandbox/github.com/apple/swift-sdk-generator/Artifacts/llvm_16.0.5.src.tar.xz
/Users/euanh/sandbox/github.com/apple/swift-sdk-generator/Artifacts/host_swift_5.9-RELEASE_x86_64-apple-macosx14.0.pkg

Unpacking and copying Swift binaries for the host triple...
xar -xf /Users/euanh/sandbox/github.com/apple/swift-sdk-generator/Artifacts/host_swift_5.9-RELEASE_x86_64-apple-macosx14.0.pkg

Time taken for this generator run: 2 seconds.
```

euanh (approved):
`Ctrl-C` now stops all the child processes.

If you cancel during the download phase the generator stops but prints at least some of the log messages for the unpacking step.   Similar things happen if you cancel during the later steps.  I'm not sure whether or not the `xar` task below was started or this was just a log message, but these messages are a little confusing.

If the user cancels we should probably not print the `Time taken for this generator run: 2 seconds.` message, or change it to `This generator run cancelled after: 2 seconds`.

```
Using downloaded artifacts in these locations:
/Users/euanh/sandbox/github.com/apple/swift-sdk-generator/Artifacts/llvm_16.0.5.src.tar.xz
/Users/euanh/sandbox/github.com/apple/swift-sdk-generator/Artifacts/host_swift_5.9-RELEASE_x86_64-apple-macosx14.0.pkg

Unpacking and copying Swift binaries for the host triple...
xar -xf /Users/euanh/sandbox/github.com/apple/swift-sdk-generator/Artifacts/host_swift_5.9-RELEASE_x86_64-apple-macosx14.0.pkg

Time taken for this generator run: 2 seconds.
```

MaxDesiatov:
> If the user cancels we should probably not print the `Time taken for this generator run: 2 seconds.` message, or change it to `This generator run cancelled after: 2 seconds`.

Unfortunately, I don't see a way to do that with the current `ServiceLifecycle` API, it doesn't rethrow `CancellationError`, so there's no way to catch it outside of `run()` where logging happens. Maybe @FranzBusch can suggest something?

FranzBusch:
> Unfortunately, I don't see a way to do that with the current `ServiceLifecycle` API, it doesn't rethrow `CancellationError`, so there's no way to catch it outside of `run()` where logging happens. Maybe @FranzBusch can suggest something?

We probably can make this configurable and let the `ServiceGroup` throw an error that indicates why it returned. Though graceful shutdown should normally not lead to something being thrown. We either make it configurable or a property on the `ServiceGroup` that can be accessed post return of `run()`

PR #53, opened 2023-11-21
Use Swift 5.9.1 and LLD 17.0.5 by default

MaxDesiatov:
These are the latest versions that we should use by default to get recent bug fixes included in produced Swift SDKs.


MaxDesiatov:
@swift-ci test

euanh (approved):

PR #54, opened 2023-11-21
Fix `fatalError` failure for development snapshots

MaxDesiatov:
Since we currently don't support `--with-docker` for development snapshots, we shouldn't attempt to calculate `swiftBaseDockerImage` when `--with-docker` is not passed. Otherwise this triggers a `fatalError` crash.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

euanh (approved):

MaxDesiatov:
@swift-ci test

PR #55, opened 2023-11-27
Rethrow error in `withEngine`

ffried:
`withEngine` currently swallows any error thrown from the closure - leading to errors being silently ignored and thus terminating the generator with a zero exit code but incomplete bundles.


MaxDesiatov (approved):
Great catch, thanks!

PR #56, opened 2023-11-27
Rethrow error from `withHTTPClient`

ffried:
In addition to the one fixed in #55, I just found another one...

This one I've found by doing a quick search for `func with` and checking the results. So all these functions that start with `with` should be fine now.
Should have done that in #55 already...

### DIFF
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Entrypoint.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Entrypoint.swift
@@ -38,6 +38,7 @@ private func withHTTPClient(
     try await client.shutdown()
   } catch {
     try await client.shutdown()
+    throw error
   }
 }
 


MaxDesiatov (approved):
Thank you! IMO we need `async defer` in the language to avoid these error-prone constructs in the future

ffried:
> Thank you! IMO we need `async defer` in the language to avoid these error-prone constructs in the future

Does that really help? AFAIK you generally can't throw errors from `defer` - and the `shutdown` calls here are throwing.

But yea, otherwise I do agree, `async defer` would be nice.

MaxDesiatov:
> Does that really help? AFAIK you generally can't throw errors from `defer` - and the `shutdown` calls here are throwing.

True, specifically for this case `async defer` is probably not applicable. I don't have a good solution for the problem formulated in my head yet, but I definitely feel that informal `with` functions convention is not good enough. At least we can abstract away `shutDown()` calls in it in the meantime, hiding the burden of clean up from the users of these closure-taking functions.

Issue #57, opened 2023-11-27
Generator fails when using docker with arm64 target arch


ffried:
The following command fails:

```bash
swift run -c release \
      swift-sdk-generator \
        --with-docker \
        --swift-version "5.9-RELEASE" \
        --host-arch "arm64" \
        --linux-distribution-name "ubuntu" \
        --linux-distribution-version "22.04" \
        --target-arch "arm64" \
        --sdk-name "test-sdk"
```

Relevant output:

```
Launching a Docker container to copy Swift SDK for the target triple from it...
                               Successfully copied 28.8MB to /Users/florianfriedrich/Projects/swift-sdk-generator/Bundles/test-sdk.artifactbundle/test-sdk/aarch64-unknown-linux-gnu/ubuntu-jammy.sdk/usr/include
Error response from daemon: Could not find the file /usr/lib64 in container 6e5fd8a034eade8c5a89991d24ce8f6913c3b2619b1111079c18f9f37430469c
6e5fd8a034eade8c5a89991d24ce8f6913c3b2619b1111079c18f9f37430469c

Unpacking and copying `lld` linker...

Fixing up absolute symlinks...

Fixing absolute paths in `glibc.modulemap`...
Error: Expected to find a file at path `/Users/florianfriedrich/Projects/swift-sdk-generator/Bundles/test-sdk.artifactbundle/test-sdk/aarch64-unknown-linux-gnu/swift.xctoolchain/usr/lib/swift/linux/aarch64/glibc.modulemap`.
```

Removing `--with-docker` leads to a successful bundle generation.

PR #58, opened 2023-11-28
Allow some paths to not exist when copying from containers

ffried:
In arm64 ubuntu images, some paths like e.g. `/usr/lib64` and `/usr/lib/arm64-linux-gnu` don't seem to exist.
I've added the possibility to ignore such failures, as well as a method to check for paths to exist in containers.

Let me know if there is a better way to check for paths in a container than how I did it.

Fixes apple/swift-sdk-generator#57.

~Also, I've found another case of a swallowed error here - which in case of #57 lead to follow-up errors later on.~
This is no longer a change in this PR - it was fixed by merging #59.

### DIFF
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
@@ -50,11 +50,11 @@ extension SwiftSDKGenerator {
         if case let containerLib64 = FilePath("/usr/lib64"),
            try await generator.doesPathExist(containerLib64, inContainer: containerID) {
           let sdkUsrLib64Path = sdkUsrPath.appending("lib64")
+          // we already checked that the path exists above, so we don't pass `failIfNotExists: false` here.
           try await generator.copyFromDockerContainer(
             id: containerID,
             from: containerLib64,
-            to: sdkUsrLib64Path,
-            failIfNotExists: false
+            to: sdkUsrLib64Path
           )
           try await createSymlink(at: pathsConfiguration.sdkDirPath.appending("lib64"), pointingTo: "./usr/lib64")
         }
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
@@ -47,29 +47,37 @@ extension SwiftSDKGenerator {
           )
         }
 
-        let sdkUsrLib64Path = sdkUsrPath.appending("lib64")
-        try await generator.copyFromDockerContainer(
-          id: containerID,
-          from: FilePath("/usr/lib64"),
-          to: sdkUsrLib64Path
-        )
-        try await createSymlink(at: pathsConfiguration.sdkDirPath.appending("lib64"), pointingTo: "./usr/lib64")
+        if case let containerLib64 = FilePath("/usr/lib64"),
+           try await generator.doesPathExist(containerLib64, inContainer: containerID) {
+          let sdkUsrLib64Path = sdkUsrPath.appending("lib64")
+          try await generator.copyFromDockerContainer(
+            id: containerID,
+            from: containerLib64,
+            to: sdkUsrLib64Path,
+            failIfNotExists: false
+          )
+          try await createSymlink(at: pathsConfiguration.sdkDirPath.appending("lib64"), pointingTo: "./usr/lib64")
+        }
 
         try await generator.createDirectoryIfNeeded(at: sdkUsrLibPath)
-        var subpaths = ["clang", "gcc", "swift", "swift_static"]
+        var subpaths: [(subpath: String, failIfNotExists: Bool)] = [
+          ("clang", true), ("gcc", true), ("swift", true), ("swift_static", true)
+        ]
 
         // Ubuntu's multiarch directory scheme puts some libraries in
         // architecture-specific directories:
         //   https://wiki.ubuntu.com/MultiarchSpec
+        // But not in all containers, so don't fail if it does not exist.
         if case .ubuntu = self.versionsConfiguration.linuxDistribution {
-          subpaths += ["\(targetTriple.cpu)-linux-gnu"]
+          subpaths += [("\(targetTriple.cpu)-linux-gnu", false)]
         }
 
-        for subpath in subpaths {
+        for (subpath, failIfNotExists) in subpaths {
           try await generator.copyFromDockerContainer(
             id: containerID,
             from: FilePath("/usr/lib").appending(subpath),
-            to: sdkUsrLibPath.appending(subpath)
+            to: sdkUsrLibPath.appending(subpath),
+            failIfNotExists: failIfNotExists
           )
         }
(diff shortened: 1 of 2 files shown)


Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
```diff
@@ -48,36 +48,42 @@ extension SwiftSDKGenerator {
...
+            // `libc.so` is a linker script with absolute paths on RHEL, replace with a relative symlink
+            let libcSO = sdkUsrLib64Path.appending("libc.so")
+            try await removeFile(at: libcSO)
+            try await createSymlink(at: libcSO, pointingTo: "libc.so.6")
+          }
         }
 
         try await generator.createDirectoryIfNeeded(at: sdkUsrLibPath)
-        var subpaths = ["clang", "gcc", "swift", "swift_static"]
+        var subpaths = [("clang", true), ("gcc", true), ("swift", true), ("swift_static", true)]
```
MaxDesiatov, line 71: Let's make clear what tuple elements stand for at the place of declaration. Alternatively, this would benefit from a separate struct that has explicitly named fields.

```suggestion
        var subpaths: [(subpath: String, failIfNotExists)] = [
          ("clang", true), ("gcc", true), ("swift", true), ("swift_static", true)
        ]
```

MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
```diff
@@ -48,36 +48,44 @@ extension SwiftSDKGenerator {
...
+            // `libc.so` is a linker script with absolute paths on RHEL, replace with a relative symlink
+            let libcSO = sdkUsrLib64Path.appending("libc.so")
+            try await removeFile(at: libcSO)
+            try await createSymlink(at: libcSO, pointingTo: "libc.so.6")
+          }
         }
 
         try await generator.createDirectoryIfNeeded(at: sdkUsrLibPath)
-        var subpaths = ["clang", "gcc", "swift", "swift_static"]
+        var subpaths: [(subpath: String, failIfNotExists)] = [
```
MaxDesiatov, line 71:
```suggestion
        var subpaths: [(subpath: String, failIfNotExists: Bool)] = [
```

MaxDesiatov: My bad, code that I previously suggested was incomplete, this should fix that.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
Just want to make sure @euanh is aware of this change as it may impact his workflows.

MaxDesiatov:
@ffried would you mind resolving conflicts?

ffried:
@MaxDesiatov Done. Thanks for the heads-up!

MaxDesiatov:
Thanks! @euanh would you have a moment to take a look at this PR? Don't want to merge this one without making sure it works well with your recent changes.

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator.swift
```diff
@@ -148,11 +148,28 @@ public actor SwiftSDKGenerator {
...
+  func doesPathExist(
+    _ containerPath: FilePath,
+    inContainer id: String
+  ) async throws -> Bool {
+    let result = try await Shell.readStdout(
+      #"\#(Self.dockerCommand) exec \#(id) sh -c 'test -e "\#(containerPath)" && echo "y" || echo "n"'"#,
+      shouldLogCommands: self.isVerbose
+    )
+    .trimmingCharacters(in: .whitespacesAndNewlines)
+    return result == "y"
```
euanh, line 160: I think you can avoid the `echo` and string comparison with something like this:
```suggestion
    do {
      try await Shell.run(
        #"\#(Self.dockerCommand) exec \#(id) sh -c 'test -e "\#(containerPath)"'"#,
        shouldLogCommands: self.isVerbose
      )
      return true
    } catch FileOperationError.nonZeroExitCode {
      return false
    }
```

euanh (approved):

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator.swift
ffried: @euanh but this would mask other errors (like if the container does not exist or does not run), right? I thought about this too, but wanted to make sure I'm not masking other errors.
Although, they might arguably show up again shortly after, or even before this.

euanh: > @euanh but this would mask other errors (like if the container does not exist or does not run), right? I thought about this too, but wanted to make sure I'm not masking other errors. Although, they might arguably show up again shortly after, or even before this.

Yes, that's true.  `doesPathExist` / `copyFromDockerContainer` is currently only ever called within a `withDockerContainer` wrapper so it should not be reached if the container failed to start, but the `docker exec` might fail if the container has crashed in the meantime.   Also, nothing in the API currently enforces this restriction - it's possible to call `copyFromDockerContainer` from outside a wrapper, in which case the provided container ID might be completely invalid.

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
```diff
@@ -47,29 +47,37 @@ extension SwiftSDKGenerator {
...
 
-        let sdkUsrLib64Path = sdkUsrPath.appending("lib64")
-        try await generator.copyFromDockerContainer(
-          id: containerID,
-          from: FilePath("/usr/lib64"),
-          to: sdkUsrLib64Path
-        )
-        try await createSymlink(at: pathsConfiguration.sdkDirPath.appending("lib64"), pointingTo: "./usr/lib64")
+        if case let containerLib64 = FilePath("/usr/lib64"),
+           try await generator.doesPathExist(containerLib64, inContainer: containerID) {
```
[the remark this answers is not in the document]
ffried: It already _is_ called with `failIfNotExists: false`... See Line 57.
Not sure where the second check is coming from...

MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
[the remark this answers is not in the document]
euanh: > It already _is_ called with `failIfNotExists: false`... See Line 57. Not sure where the second check is coming from...
> Ah, it actually needs to be failIfNotExists: true now to avoid the second check...

Oops, I got it the wrong way around in my comment - I should have said `failIfNotExists: true`. 🙃

MaxDesiatov:
@ffried apologies for the delay with merging this. Unfortunately, I can't push to the PR branch, would you mind merging `main` into it, which will unblock merging for me? Thank you!

ffried:
@MaxDesiatov no worries! Rebase is done.

MaxDesiatov:
@swift-ci test

PR #59, opened 2023-11-28
Add (async) throwing defer helpers

ffried:
This offers a more generalized solution to the bugs fixed in #55 and #56 by adding a `Helpers` target that (currently) contains `withThrowingDefer` and `withAsyncThrowingDefer`.

Those functions basically do what was previously done manually at each such occasion:
- Run a given `work` closure
- Run a given `deferred` closure no matter what
- Re-throw errors from `work` or `deferred` (where errors from the latter take precedence).

Also, a new `withDockerContainer` function was added (fixes the swallowed error, which I also fixed in #58).

### DIFF
--- a/Sources/GeneratorEngine/Engine.swift
+++ b/Sources/GeneratorEngine/Engine.swift
@@ -14,6 +14,7 @@ import class AsyncHTTPClient.HTTPClient
 @_exported import Crypto
 import struct Logging.Logger
 @_exported import struct SystemPackage.FilePath
+import Helpers
 
 public func withEngine(
   _ fileSystem: any FileSystem,
@@ -27,12 +28,10 @@ public func withEngine(
     cacheLocation: cacheLocation
   )
 
-  do {
+  try await withAsyncThrowingDefer {
     try await body(engine)
+  } deferring: {
     try await engine.shutDown()
-  } catch {
-    try await engine.shutDown()
-    throw error
   }
 }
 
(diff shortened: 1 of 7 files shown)


MaxDesiatov:
@weissi I'm curious if you have an opinion on this?

FranzBusch:
FWIW, I have implemented similar methods before but the one difference is that I (and @weissi) have put the deferred closure into an unstructured task `Task {}`. The reason for this is that often the code in the defer **must** run but if the task is already cancelled that code will also get cancelled. I haven't looked at the usages here but this might be an important thing to consider.

ffried:
@FranzBusch That is a very good point! It this case, it does not make things worse (since the previous implementations actually suffered from the same problem), but it's still something worth having a look at.

With these new helper functions, we can at least change it in a central place (vs. having to adjust multiple sites that follow this pattern).

MaxDesiatov:
@swift-ci test

Sources/GeneratorEngine/Engine.swift
```diff
@@ -27,12 +28,10 @@ public func withEngine(
     cacheLocation: cacheLocation
   )
 
-  do {
+  try await withAsyncThrowingDefer {
     try await body(engine)
+  } deferring: {
```
MaxDesiatov, line 33: WDYT of a slightly different naming?

```suggestion
  try await withAsyncThrowing {
    try await body(engine)
  } defer: {
```

ffried: It seems to build. Xcode (SourceKit) will format the `defer` here like the actual `defer` keyword, but otherwise it seems to work.

MaxDesiatov:
@swift-ci test

MaxDesiatov (approved):
Thanks!

PR #66, opened 2024-01-11
Introduce `SwiftSDKRecipe` to abstract platform-specific code

kateinoigakukun:
A recipe should describe how to package a whole SDK and a SDK generator should provide a way to generate a part of the SDK and used by the recipe.

I'm going to get rid of `LinuxDistribution` from `SwiftSDKGenerator` actor and move all uses of it to `LinuxRecipe`

### DIFF
--- a/Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
+++ b/Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
@@ -1,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2022-2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import AsyncHTTPClient
+import GeneratorEngine
+
+public struct LinuxRecipe: SwiftSDKRecipe {
+  public init() {}
+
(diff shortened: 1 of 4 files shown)


kateinoigakukun:
@swift-ci Please test

Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import AsyncHTTPClient
+import GeneratorEngine
+
+public struct LinuxRecipe: SwiftSDKRecipe {
+  public init() {}
+
+  public func cookSwiftSDK(generator: SwiftSDKGenerator, engine: Engine, httpClient client: HTTPClient) async throws {
```
MaxDesiatov, line 19:
```suggestion
  public func prepareSwiftSDK(generator: SwiftSDKGenerator, engine: Engine, httpClient client: HTTPClient) async throws {
```

MaxDesiatov (commented):
This is a great start in that direction, thank you so much! My only suggestion here is maybe not to take the recipes analogy too literally and use `prepare` instead of `cook`? 🧑‍🍳

Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
kateinoigakukun: I feel `prepare` implies that the operation is not the main part but a minor part and subsequent operations are the main. How about `make`?

MaxDesiatov: Yes, that makes sense, and I know there's a lot of precedent for `make` in swift APIs and also with build systems, I like `make` 👍

kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #72, opened 2024-01-24
Use SwiftDriver's `Triple` parser

kateinoigakukun:
SwiftDriver's `Triple` parser is more robust than our bespoke one, and it's already tested.

* Replace all use of `Triple` with the SwiftDriver's one
* "uname -m" and "sw_vers" commands are replaced with libc `uname`
* Darwin version -> macOS version conversion is now done by the `Triple`


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):
LGTM

kateinoigakukun:
@swift-ci test

PR #73, opened 2024-01-24
Replace `--target-arch/--host-arch <arch>` with `--target/--host <triple>`

kateinoigakukun:
This allows users to specify other triple fields like os and vendor. It's NFC at this moment but will be used to distinguish `wasm32-unknown-wasi` and `wasm32-unknown-none`

### DIFF
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -100,28 +100,21 @@ extension GeneratorCLI {
 
     @Option(
       help: """
-      CPU architecture of the host triple of the bundle. Defaults to a triple of the machine this generator is \
-      running on if unspecified. Available options: \(
-        Triple.Arch.allCases.map { "`\($0.rawValue)`" }.joined(separator: ", ")
-      ).
+      The host triple of the bundle. Defaults to a triple of the machine this generator is \
+      running on if unspecified.
       """
     )
-    var hostArch: Triple.Arch? = nil
+    var host: Triple? = nil
 
     @Option(
       help: """
-      CPU architecture of the target triple of the bundle. Same as the host triple CPU architecture if unspecified. \
-      Available options: \(Triple.Arch.allCases.map { "`\($0.rawValue)`" }.joined(separator: ", ")).
+      The target triple of the bundle. Same as the host triple if unspecified.
       """
     )
-    var targetArch: Triple.Arch? = nil
+    var target: Triple? = nil
 
     func deriveHostTriple() throws -> Triple {
-      let current = try SwiftSDKGenerator.getCurrentTriple(isVerbose: verbose)
-      if let explicitArch = hostArch {
-        return Triple(arch: explicitArch, vendor: current.vendor, os: current.os!)
-      }
-      return current
+      try host ?? (try SwiftSDKGenerator.getCurrentTriple(isVerbose: verbose))
     }
   }
 
@@ -161,7 +154,7 @@ extension GeneratorCLI {
     var linuxDistributionVersion: String?
 
     func deriveTargetTriple(hostTriple: Triple) -> Triple {
-      Triple(arch: self.generatorOptions.targetArch ?? hostTriple.arch!, vendor: nil, os: .linux, environment: .gnu)
+      self.generatorOptions.target ?? Triple(arch: hostTriple.arch!, vendor: nil, os: .linux, environment: .gnu)
     }
 
     func run() async throws {
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -108,7 +108,7 @@ extension GeneratorCLI {
 
     @Option(
       help: """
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -100,28 +100,21 @@ extension GeneratorCLI {
...
       """
     )
-    var hostArch: Triple.Arch? = nil
+    var host: Triple? = nil
 
     @Option(
       help: """
-      CPU architecture of the target triple of the bundle. Same as the host triple CPU architecture if unspecified. \
-      Available options: \(Triple.Arch.allCases.map { "`\($0.rawValue)`" }.joined(separator: ", ")).
+      The target triple of the bundle. Same as the host triple if unspecified.
```
MaxDesiatov, line 111: This updated help string looks wrong to me, unless the SDK generator doesn't generate anything useful in its default configuration? What SDK does it generate if host and target triples are the same and you run it on macOS?

kateinoigakukun: Oh, good catch. The default target triple now depends on the recipe subcommand, how should we describe it in help 🤔

euanh: I think the previous behaviour (when the generator was mainly for Linux SDKs) was to match the host and target CPU architecture if CPU wasn't specified.

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -100,35 +100,31 @@ extension GeneratorCLI {
...
 
     @Option(
       help: """
-      CPU architecture of the target triple of the bundle. Same as the host triple CPU architecture if unspecified. \
-      Available options: \(Triple.Arch.allCases.map { "`\($0.rawValue)`" }.joined(separator: ", ")).
+      The target triple of the bundle. The default depends on a recipe used for SDK generation. Pass `--help` to a specific recipe subcommand for more details.
       """
     )
-    var targetArch: Triple.Arch? = nil
+    var target: Triple? = nil
```
MaxDesiatov, line 114: Do we need to keep previous option names for compatibility? I'm not sure if @euanh is strictly relying on the current spelling in CLI invocations or on the higher level API.

MaxDesiatov: @kateinoigakukun I guess if it's invoked as `swift-sdk-generator make-linux-sdk` nothing would break, right?

euanh: It looks like the `--target/--host` triple arguments are in the top level `swift-sdk-generator` call.  Maybe they should be delegated to the subcommand?
Alternatively, if we have the full `--target` triple, we have all the information we need to dispatch to Linux or WASI SDK generation, so maybe we don't need the `make-linux-sdk/make-wasi-sdk` commands at all?

kateinoigakukun: I'm ok to keep that compatibility 👍 

@MaxDesiatov 
> I guess if it's invoked as swift-sdk-generator make-linux-sdk nothing would break, right?

Ah, good idea. Currently `swift-sdk-generator` is just an alias of `swift-sdk-generator make-linux-sdk`, so they both don't have compatibility with old one. But we can keep compatibility only for the non-subcommand version by having `GeneratorCLI.run()` separately.

@euanh 
> It looks like the --target/--host triple arguments are in the top level swift-sdk-generator call. Maybe they should be delegated to the subcommand?

The `GeneratorOptions` that `--target/--host` belongs to is embedded for each subcommand with `@OptionGroup`, so it's already delegated to subcommand.


> Alternatively, if we have the full --target triple, we have all the information we need to dispatch to Linux or WASI SDK generation, so maybe we don't need the make-linux-sdk/make-wasi-sdk commands at all?

We still need those subcommand separation to have different set of options for each recipe subcommand

MaxDesiatov: I don't care too much for compatibility at this point since we haven't tagged a single version yet. But I want to make sure these changes don't breaking anything for Euan unnecessarily, or at least we communicate them clearly and there's a clear migration path.

kateinoigakukun: Makes sense to me. Updated to keep compatibility while emitting deprecation warning

MaxDesiatov (approved):
With preserved compatibility and a deprecation/migration path this LGTM. Thanks!

MaxDesiatov:
@swift-ci test

kateinoigakukun:
@swift-ci test

euanh (approved):

Sources/GeneratorCLI/GeneratorCLI.swift
euanh: @kateinoigakukun Could you update the PR title to reflect that it is just deprecating the old options?
I tweaked it slightly because it used to say `--target-cpu` but the flag is actually `--target-arch`.

kateinoigakukun:
@swift-ci test

PR #77, opened 2024-01-25
Update `AsyncProcess`, use it in `Shell` implementation

MaxDesiatov:
These changes make `AsyncProcess` implementation more robust and increase its code coverage.

Now we no longer have to rely on `Foundation.Process` directly in `Shell`, as `AsyncProcess` abstracts rough edges away for us and also provides an async-friendly API.


MaxDesiatov:
@swift-ci test

PR #80, opened 2024-01-28
Remove unused toolchain components from WebAssembly SDK

kateinoigakukun:
Currently, Wasm SDK contains whole host toolchain package contents, but some components are unnecessary and duplicated with the installed host toolchain.  This PR removes unused components as Linux SDK generator does, and in addition, removes lldb-related components.

The total reduction in `.artifactbundle` for Linux host is: 2.7GB -> 1.3GB ( -1.1GB by platform libraries and host tools, -300MB by lldb components)

I don't know if we should remove lldb components also in Linux SDK generator 🤔

### DIFF
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Unpack.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Unpack.swift
@@ -13,7 +13,7 @@
 import GeneratorEngine
 import struct SystemPackage.FilePath
 
-private let unusedDarwinPlatforms = [
+let unusedDarwinPlatforms = [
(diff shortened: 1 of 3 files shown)


MaxDesiatov (approved):
How useful is lldb in a Wasm SDK? I reckon one would always use the host lldb for anything host-related, since there's no Wasm as a target platform in lldb .

MaxDesiatov:
@swift-ci test

kateinoigakukun:
At least for now, we don't have any Swift specific use case of lldb as you said, so not so much useful.

MaxDesiatov:
For the sake of keeping PRs small, I suggest removing LLDB in a separate PR though.

PR #82, opened 2024-02-02
Keep Swift libraries linked by compiler tools

kateinoigakukun:
Swift libraries for host platforms are linked by the Swift compiler tools, so they should not be removed from the toolchain directory.

I thought keeping `lib/host` is enough, but `swift-frontend` also links `../lib/linux/libswiftCore.so` on Linux host.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):
Good catch, thank you 👍

PR #85, opened 2024-02-07
WebAssembly: Specify resource directory explicitly

kateinoigakukun:
When compiler is vendored in the SDK, the resource directory is inferred from the location of the compiler executable. This is not the case when the compiler is not vendored, so we need to specify the resource directory explicitly to work with both cases.

### DIFF
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Metadata.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Metadata.swift
@@ -69,7 +69,8 @@ extension SwiftSDKGenerator {
     )
 
     recipe.applyPlatformOptions(
-      metadata: &metadata, paths: pathsConfiguration,
+      metadata: &metadata,
+      paths: pathsConfiguration,
       targetTriple: self.targetTriple
     )
 
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Metadata.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Metadata.swift
@@ -43,7 +43,7 @@ extension SwiftSDKGenerator {
     return toolsetJSONPath
   }
 
-  func generateDestinationJSON(toolsetPath: FilePath, sdkDirPath: FilePath) throws {
+  func generateDestinationJSON(toolsetPath: FilePath, sdkDirPath: FilePath, recipe: SwiftSDKRecipe) throws {
     logGenerationStep("Generating destination JSON file...")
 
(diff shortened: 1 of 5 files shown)


Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Metadata.swift
```diff
@@ -63,15 +63,22 @@ extension SwiftSDKGenerator {
       """)
     }
 
+    var metadata = SwiftSDKMetadataV4.TripleProperties(
+      sdkRootPath: relativeSDKDir.string,
+      toolsetPaths: [relativeToolsetPath.string]
+    )
+
+    recipe.applyPlatformOptions(
+      metadata: &metadata, paths: pathsConfiguration,
```
MaxDesiatov, line 72: formatting nit
```suggestion
      metadata: &metadata,
      paths: pathsConfiguration,
```

Sources/SwiftSDKGenerator/Serialization/SwiftSDKMetadata.swift
```diff
@@ -73,8 +73,8 @@ struct DestinationV3: Encodable {
 }
 
 /// Represents v4 schema of `swift-sdk.json` (previously `destination.json`) files used for cross-compilation.
-struct SwiftSDKMetadataV4: Encodable {
-  struct TripleProperties: Encodable {
+public struct SwiftSDKMetadataV4: Encodable {
+  public struct TripleProperties: Encodable {
```
MaxDesiatov, line 77: Let's avoid `public` in new code and make all APIs explicitly local to this package.

```suggestion
package struct SwiftSDKMetadataV4: Encodable {
  package struct TripleProperties: Encodable {
```

Sources/SwiftSDKGenerator/SwiftSDKRecipes/SwiftSDKRecipe.swift
```diff
@@ -32,4 +37,9 @@ public protocol SwiftSDKRecipe: Sendable {
 
 extension SwiftSDKRecipe {
   public func applyPlatformOptions(toolset: inout Toolset) {}
+  public func applyPlatformOptions(
```
MaxDesiatov, line 40:
```suggestion
  package func applyPlatformOptions(toolset: inout Toolset) {}
  package func applyPlatformOptions(
```

kateinoigakukun:
@swift-ci test

MaxDesiatov:
Ah, sorry, looks like `package` needs to be applied in a few more places. Feel free to revert that commit and we'll apply the conversion for use of `package` on the whole project at once in a separate PR.

kateinoigakukun:
@swift-ci test

PR #86, opened 2024-02-07
WebAssembly: Copy clang headers from target toolchain

kateinoigakukun:
The clang headers were copied from host toolchain at the first "rsync" step. But host toolchain can be optional, so we should copy the headers from the target toolchain instead.


MaxDesiatov (approved):
Thanks!

MaxDesiatov:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

PR #91, opened 2024-03-29
Make `supportedTriples` optional in SDK metadata for universal SDKs

kateinoigakukun:
For SDKs that can be universally used on any host platform, the `supportedTriples` field in the SDK metadata was previously required to fill all possible host triples.
This change makes the `supportedTriples` field optional, allowing SDKs to be universally used without specifying any host triples, and Wasm SDK generator fills the field with null when `--host-swift-package-path` is not provided.

We need SwiftPM side support as well. https://github.com/apple/swift-package-manager/pull/7432

### DIFF
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -26,7 +26,6 @@ struct GeneratorCLI: AsyncParsableCommand {
 
   static func run<Recipe: SwiftSDKRecipe>(
     recipe: Recipe,
-    hostTriple: Triple,
     targetTriple: Triple,
     options: GeneratorOptions
   ) async throws {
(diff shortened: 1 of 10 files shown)


MaxDesiatov (approved):
Thanks!

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -26,15 +26,13 @@ struct GeneratorCLI: AsyncParsableCommand {
 
   static func run<Recipe: SwiftSDKRecipe>(
     recipe: Recipe,
```
MaxDesiatov, line 28: Could take this as an opportunity to simplify the generic signature 🙂 
```suggestion
  static func run(
    recipe: some SwiftSDKRecipe,
```

kateinoigakukun:
@swift-ci test

PR #92, opened 2024-03-29
Do not create a symlink for `swift-autolink-extract` if `swift` is not installed

kateinoigakukun:
If `swift` is not installed in the final toolchain in the SDK, we don't need `swift-autolink-extract` either.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

PR #96, opened 2024-04-04
Remove swift-service-lifecycle dependency

kateinoigakukun:
We have only single service in the project, so we don't need to orchestrate multiple services to gracefully shutdown. We can just run the single service and let it run until it finishes.

### DIFF
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -12,7 +12,6 @@
 
 import ArgumentParser
 import Logging
-import ServiceLifecycle
 import SwiftSDKGenerator
 import struct SystemPackage.FilePath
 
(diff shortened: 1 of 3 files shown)


Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -40,18 +39,7 @@ struct GeneratorCLI: AsyncParsableCommand {
         logger: logger
       )
 
-      let serviceGroup = ServiceGroup(
-        configuration: .init(
-          services: [.init(
-            service: SwiftSDKGeneratorService(recipe: recipe, generator: generator),
-            successTerminationBehavior: .gracefullyShutdownGroup
-          )],
-          cancellationSignals: [.sigint],
```
MaxDesiatov, line 49: We need to reimplement functionality introduced by `cancellationSignals: [.sigint]` if this is to be removed, otherwise Ctrl-C in terminal on a running generator process doesn't send signals to child processes. I assume corresponding code in `swift-service-lifecycle` is self-contained and we can just vendor it here?

MaxDesiatov, line 49: We need to reimplement functionality introduced by cancellationSignals: [.sigint] if this is to be removed, otherwise Ctrl-C in terminal on a running generator process doesn't cancel ongoing AsyncProcess tasks, which would send signals to child processes. I assume corresponding code in swift-service-lifecycle is self-contained and we can just vendor it here?

MaxDesiatov (changes requested):
We need to reimplement functionality introduced by `cancellationSignals: [.sigint]` if this is to be removed, otherwise Ctrl-C in terminal on a running generator process doesn't cancel ongoing AsyncProcess tasks, which would send signals to child processes. I assume corresponding code in swift-service-lifecycle is self-contained and we can just vendor it here?

FranzBusch:
What's the motivation here to remove the dependency? The code over in service lifecycle does a few things to handle the signals. Furthermore, it also has built in protection for when cancellation isn't happening in a certain time frame to force a crash (This is opt-in). I would love to understand why we want to get rid of the dependency

FranzBusch:
If you would want to vendor it then you probably want to pull in the `UnixSignals` code which does all the work to bring signals into Swift Concurrency and then you can setup a simple task group that listens for `SIGINT` and cancels the actual work.

kateinoigakukun:
On second thoughts, it might be reasonable to add it to the checkout list?

MaxDesiatov:
> On second thoughts, it might be reasonable to add it to the checkout list?

That's a separate process and not sure how much time it would take if considered in the first place.

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #97, opened 2024-04-04
Fix `noCacheKeyConformance` error on `make-linux-sdk`

kateinoigakukun:
Follow-up to 3e076d11485793efc1fde511eb89d14a76d9b420

SystemString used inside FilePath is not a CacheKey, so the encoder complains about it, but we can consider FilePath as a leaf type and no need to encode it further.


MaxDesiatov:
@swift-ci test

PR #98, opened 2024-04-05
Fix undeterministic hash key due to pointer address in anonymous context name

kateinoigakukun:
`CodingKeys` is automatically derived and defined as `private`, and types defined as private have anonymous context in its mangled name. Then `String(reflecting:)` returns the qualified name of the type including the anonymous context name, which contains the pointer address like `Item.(unknown context at $55eab1620198).CodingKeys` and the address can be different across runs due to ASLR.

So we need to stop encoding CodingKeys to reduce the possibility of undeterministic hash key. Our hashing strategy can be still underteministic if the cache key itself is declared as private, but we can't avoid it while keeping the cache key nominal type sensitive (not structurally sensitive).

### DIFF
--- a/Tests/GeneratorEngineTests/EngineTests.swift
+++ b/Tests/GeneratorEngineTests/EngineTests.swift
@@ -149,4 +149,28 @@ final class EngineTests: XCTestCase {
 
     try await engine.shutDown()
   }
+
+  struct MyItem: Sendable, CacheKey {
+    let remoteURL: URL
+    var localPath: FilePath
+    let isPrebuilt: Bool
+  }
+
+  func testQueryEncoding() throws {
+    let item = MyItem(
+      remoteURL: URL(string: "https://download.swift.org/swift-5.9.2-release/ubuntu2204-aarch64/swift-5.9.2-RELEASE/swift-5.9.2-RELEASE-ubuntu22.04-aarch64.tar.gz")!,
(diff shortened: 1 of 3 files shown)


kateinoigakukun:
@swift-ci test

Tests/GeneratorEngineTests/EngineTests.swift
```diff
@@ -149,4 +149,35 @@ final class EngineTests: XCTestCase {
...
+
+  struct MyItem: Sendable, CacheKey {
+    let remoteURL: URL
+    var localPath: FilePath
+    let isPrebuilt: Bool
+    init(remoteURL: URL, localPath: FilePath, isPrebuilt: Bool) {
+        self.remoteURL = remoteURL
+        self.localPath = localPath
+        self.isPrebuilt = isPrebuilt
+    }
```
MaxDesiatov, line 161: IIUC this is equivalent to the synthesized default initializer and is not needed?
```suggestion
```

MaxDesiatov (approved):
LGTM, just a small nit

Tests/GeneratorEngineTests/EngineTests.swift
kateinoigakukun: Ah, that was just for debugging purpose

kateinoigakukun:
@swift-ci test

PR #99, opened 2024-04-05
Send SIGINT to the process group instead of SIGKILL to the process itself

kateinoigakukun:
The executor was sending SIGKILL to the process itself, which caused orphaned processes to be left when multiple processes were spanwed by the process (e.g. shell). This change sends SIGINT to the process group instead, which lets the shell handle the signal and propagate it to its children.

### DIFF
--- a/Sources/AsyncProcess/ProcessExecutor.swift
+++ b/Sources/AsyncProcess/ProcessExecutor.swift
@@ -486,7 +486,7 @@ public final actor ProcessExecutor {
       }
       if p.isRunning {
         self.logger.info("terminating process", metadata: ["pid": "\(childPid)"])
-        kill(childPid, SIGKILL)
+        kill(-childPid, SIGINT)
       } else {
         self.logger.debug("child process already dead", metadata: ["pid-if-available": "\(childPid)"])
       }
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

Sources/AsyncProcess/ProcessExecutor.swift
```diff
@@ -486,7 +486,7 @@ public final actor ProcessExecutor {
       }
       if p.isRunning {
         self.logger.info("terminating process", metadata: ["pid": "\(childPid)"])
-        kill(childPid, SIGKILL)
+        kill(-childPid, SIGINT)
```
weissi, line 489: I wouldn't recommend that. Structured Concurrency requires that the process _will definitely exit_ or else we're just gonna deadlock. But I think we can update `ProcessExecutor` to be able to run a shutdown sequence instead of just sending SIGKILL

kateinoigakukun: Sorry, does "process" mean spawned child process, or the SDK generator process itself? I don't think `kill(2)` does not block the sender process.

weissi: `kill` doesn't block. But let's assume that the child doesn't exit if you send a `kill(SIGINT, child)`, then we'll potentially wait for that child forever. So it's very important that (at some point) we do send a `SIGKILL` if the child doesn't disappear by itself.

But see #100 for teardown sequences which should fix this

kateinoigakukun: Ah, that sounds reasonable. Thanks!

PR #103, opened 2024-04-14
Remove `linuxConventionDescription` conversion

kateinoigakukun:
Now SDK generator users are responsible for providing appropriate triples for their targets, so we don't need to convert them to the Linux convention anymore.

Also stop normalizing the triple string given through CLI options to keep the original triple string as is in the generated SDK metadata. This makes it more flexible to support various triple strings like wasm32-unknown-wasip1-threads.


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test macOS

MaxDesiatov (approved):

kateinoigakukun:
Thanks!

PR #119, opened 2024-08-08
Downgrade .swiftformat to Swift 5.8

MaxDesiatov:
Compatibility with Swift 5.8 should be maintained for as long as we're building with Swift 5.8 on CI.


MaxDesiatov:
@swift-ci test

PR #122, opened 2024-08-27
Bump dependencies to match `update-checkout-config.json`

MaxDesiatov:


euanh:
`testCanDealWithRunawayChildProcesses` failed on macOS but it does not seem to be caused by this change - that test has been failing for several builds.

PR #126, opened 2024-08-27
Reduce Linux builds verbosity in `test.sh`

MaxDesiatov:
`-vv` makes logs too large and hard to read. We don't need that logging level right now and can easily bring it back when needed.


MaxDesiatov:
@swift-ci test

PR #127, opened 2024-08-27
Adopt `swift-toolchain-sqlite` to reduce system dependencies

MaxDesiatov:
This means we no longer require users to provide their own system SQLite installation on Linux or Windows.


MaxDesiatov:
@swift-ci test

bnbarham:
Suppose this is a question for @jakepetroules - should swift-toolchain-sqlite use the system dependency for Apple platforms, or is it intentional that all platforms are just using the amalgamation?

MaxDesiatov:
@swift-ci test

MaxDesiatov:
FWIW without this change due to SwiftPM's bug in `systemLibrary` support we get this warning on every macOS build
```
warning: failed to retrieve search paths with pkg-config; maybe pkg-config is not installed
warning: couldn't find pc file for sqlite3
```

MaxDesiatov:
@swift-ci test

xtremekforever:
Are you going to merge this? Just curious, as it seems like it could be a good improvement.

MaxDesiatov:
@swift-ci test

jakepetroules:
> Suppose this is a question for @jakepetroules - should swift-toolchain-sqlite use the system dependency for Apple platforms, or is it intentional that all platforms are just using the amalgamation?

Sorry I'm only seeing this now -- it's the intention that only platforms which need it are using the amalgamation, while platforms that have an accessible system dependency use that one. So amalgamation would currently be Windows and WebAssembly IIRC.

bnbarham:
> See llbuild for example of usage.

I was more wondering whether we think `swift-toolchain-sqlite` should do this itself (ie. have an empty target that depends on the system sqlite in the non-windows/wasm case), rather than all clients in the toolchain.

MaxDesiatov:
Swift SDK Generator is not a part of the installable toolchain, although it's a part of some toolchain CI jobs. Swift SDK Generator users build it locally on their machine vast majority of the time, and minimizing the amount of prerequisites for them is crucial. That's why I'd prefer to use this new package on Linux too, especially SQLite dev headers installation instructions can be so different for every Linux distribution.

PR #129, opened 2024-08-28
WasmSDK: Copy swift-foundation headers and apinotes into the Swift SDK bundle

kateinoigakukun:
Those files are recently added to the toolchain, and they are required to build Swift code that uses Foundation on WebAssembly.

We had been manually specifying the paths to copy from the target package into the SDK, but we now build host-agnostic SDKS. So we can just copy the whole `usr/lib` from the target Swift package into the SDK bundle instead of adding individual new paths. The host-specific build path is still used for 5.10 SDKs, so keep the old code as is for now, but it will be removed in the future.


kateinoigakukun:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test macos

MaxDesiatov:
@swift-ci test

kateinoigakukun:
Verified in https://github.com/swiftwasm/swiftwasm-build/pull/418

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

PR #132, opened 2024-08-28
Address Foundation-related FIXME in `GeneratorCLI.swift`

MaxDesiatov:
Now that `Duration.formatted()` is available on non-Darwin platforms in Swift 6.0+ we can utilize it in `GeneratorCLI.swift` and address a long-standing `FIXME`.


MaxDesiatov:
@swift-ci test

Issue #76, opened 2024-01-24
Replace custom `async` file system wrapper with `NIOFileSystem` 


MaxDesiatov:
We currently have our own, not necessarily well-tested and valid implementation of `async` file system APIs in https://github.com/apple/swift-sdk-generator/tree/main/Sources/GeneratorEngine/FileSystem. With `NIOFileSystem` now available in SwiftNIO 2.63, we should use it instead.

zaneenders:
Hi, I was wondering if I could work on this? 

As I was poking around I noticed that running `./Utilities/soundness.sh` on main fails for `UnixName.swift`
```
=> Checking license headers... 
   * swift-or-c... missing headers in file './Sources/SwiftSDKGenerator/SystemUtils/UnixName.swift'!
--- /dev/fd/63  2024-08-03 15:39:55
+++ /tmp/.swift-package-manager-soundness_nooVYa        2024-08-03 15:39:55
```

Also this is my first time using `swiftformat` but when running `swiftformat .` on main, 35 files are reformatted.
```
➜  swift-sdk-generator git:(main) swiftformat .
Running SwiftFormat...
Reading config file at .../swift-sdk-generator/.swiftformat
warning: --varattributes option is deprecated. Use with `--storedvarattrs` or `--computedvarattrs` instead.
SwiftFormat completed in 0.46s.
34/60 files formatted.
```

MaxDesiatov:
Yes, as this is an issue marked as "good first issue" and is unassigned, it's free for anyone for anyone to work on.

Feel free to fix soundness and formatting issues, but those should be submitted as separate PRs so that those changes don't pollute other PRs one may submit as part of their work.

MaxDesiatov:
It won't change the status quo WRT Window support, as Swift SDK Generator already depends on SwiftNIO.

compnerd:
Ah, that's rather unfortunate.

zaneenders:
Could the new Foundation help with the NIO dependency? Though it doesn't look like SubProcess has made it in yet.
https://github.com/apple/swift-foundation/pull/439

MaxDesiatov:
No, it won't help since relevant Foundation APIs are not `async`.

PR #137, opened 2024-09-30
Use _NIOFileSystem for file system operations.

zaneenders:
Motivation:
Add support for _NIOFileSystem to move away from OSFileSystem as per issue #76.

Modifications:
Add SDKFileSystem that uses _NIOFileSystem. Change default AsyncFileSystem to SDKFileSystem

Result:

Using Swift NIO for file system operations.

### DIFF
--- a/Sources/Helpers/Vendor/_AsyncFileSystem/SDKFileSystem.swift
+++ b/Sources/Helpers/Vendor/_AsyncFileSystem/SDKFileSystem.swift
@@ -1,0 +1,59 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift open source project
+//
+// Copyright (c) 2023-2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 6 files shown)


MaxDesiatov:
@swift-ci test

Sources/Helpers/Vendor/_AsyncFileSystem/SDKFileSystem.swift
```diff
@@ -0,0 +1,59 @@
...
+            guard let _ = try await FileSystem.shared.info(forFileAt: path) else {
+                return false
+            }
+            return true
+        } catch {
+            return false
+        }
+    }
+
+    public func withOpenReadableFile<T: Sendable>(
```
MaxDesiatov, line 32: I don't think we should reinvent the wheel here and reimplement `with*` methods from scratch. The main motivation for using `_NIOFileSystem` in the first place is that it already [implements these methods](https://swiftpackageindex.com/apple/swift-nio/2.73.0/documentation/_niofilesystem/filesystem/withfilehandle(forreadingandwritingat:options:execute:)) and provides necessary abstractions, like [`FileSystemProtocol`](https://swiftpackageindex.com/apple/swift-nio/2.73.0/documentation/_niofilesystem/filesystemprotocol).

MaxDesiatov (changes requested):

zaneenders:
Hey, some personal matters have come up and I don't think I will be able to work on and finish this in a timely matter. I'm sorry for taking so long to notify you.

PR #158, opened 2024-11-27
Fix 5.8 compatibility

kateinoigakukun:
We still use 5.8 as a bootstrap compiler on CI, so we need to make sure that the code is compatible with it.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

euanh:
@kateinoigakukun @MaxDesiatov Do we have a CI job somewhere which checks for 5.8 compatibility?

MaxDesiatov:
Not on this repo directly, but this job is impacted https://ci.swift.org/job/swift-PR-Linux-crosscompile-wasm/ if 5.8 compatibility is broken.

Issue #65, opened 2024-01-10
armhf support


lhoward:
How difficult would it be to add `armhf` target architecture support (for Linux)?

MaxDesiatov:
This would definitely require a Swift toolchain build, which I'm actually interested in supporting in the generator eventually.

Are you targeting a specific Linux distribution? Would you be able to clarify what that is and its exact version for development and testing?

lhoward:
I'm building on Ubuntu 22.04.3 `aarch64-linux-gnu` and would like to target the same distribution, but on `arm-linux-gnueabihf`. (I'm working on an embedded product and I'm keen to see if I can save a bit on BOM costs but using a 32-bit CPU.)

xtremekforever:
I'm also interested in this but have been unable to build a toolchain for Swift for armhf. We have been able to build llvm, swift, and the core libraries for armhf to be able to put together a simple sysroot for Debian 11 and 12, but beyond that that we don't have any other tools built.

I wonder how much of a toolchain we actually need to support armv7

MaxDesiatov:
Do you have a preset for that platform defined in the presets file? https://github.com/swiftlang/swift/blob/main/utils/build-presets.ini

xtremekforever:
I don't have anything in build-presents.ini, instead we've been using cmake due to the fact that it's easier to customize and due to a small bug in the swift-build-support build scripts when cross compiling for armv7. But, I could give it another try. Is it okay for a build preset to be defined to cross compile for armv7 and still work for the generator?

MaxDesiatov:
Everything buildable with a preset locally can be built with a buildbot too, I'm not aware of any differences between those two cases that would preclude cross-compilation.

xtremekforever:
In my experience using build presets with buildbot to try to cross compile armv7 is a pain. There's always an issue of LLVM not finding something, some path being wrong, or it straight up doing a step that doesn't work for cross compilation. I will continue looking at it, but I wonder if the generator would work if I just pointed it to some artifacts that were built with cmake that are for armv7...which would include Swift stdlib, dispatch, foundation, and xctest.

xtremekforever:
[...]

PR #170, opened 2025-01-21
Add support for building Swift SDKs for armv7 with target Swift toolchain or docker container

xtremekforever:
This would implement #65. Although there is no official support for armv7 in Swift yet, this opens the door for being able to generate Swift SDKs for cross compilation to armv7 given you have build artifacts to give it. In my case, I can grab some that I built from my [swift-armv7 repo](https://github.com/xtremekforever/swift-armv7/releases/tag/6.0.3), download and extract, then run the generator like this:

```bash
$ swift run swift-sdk-generator make-linux-sdk --swift-version 6.0.3-RELEASE --target armv7-unknown-linux-gnueabihf --target-swift-package-path ~/Downloads/swift-6.0.3-RELEASE-ubuntu-jammy-armv7-install
...
All done! Install the newly generated SDK with this command:
swift experimental-sdk install ~/swift-sdk-generator/Bundles/6.0.3-RELEASE_ubuntu_jammy_armv7.artifactbundle

swift build --experimental-swift-sdk 6.0.3-RELEASE_ubuntu_jammy_armv7
```

And, this generates a working Swift SDK that I can use to cross compile anything, like a Hummingbird app:

```bash
$ swift build --swift-sdk 6.0.3-RELEASE_ubuntu_jammy_armv7
warning: 'swift-algorithms': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    ~/hummingbird-examples/hello/.build/checkouts/swift-algorithms/Sources/Algorithms/Documentation.docc
Building for debugging...
[948/948] Linking App
Build complete! (35.16s)
```

So, I think that it could be reasonable to add this support to the generator now. It's something I will use myself, but also when official support for armv7 lands for Swift, it'll be ready ;)

### DIFF
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
@@ -73,6 +73,11 @@ extension SwiftSDKGenerator {
         // But not in all containers, so don't fail if it does not exist.
         if case .ubuntu = targetDistribution {
           subpaths += [("\(targetTriple.archName)-linux-gnu", false)]
+
+          // Custom subpath for armv7
+          if targetTriple.archName == "armv7" {
+            subpaths += [("arm-linux-gnueabihf", false)]
+          }
         }
 
         for (subpath, failIfNotExists) in subpaths {
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
@@ -109,13 +109,16 @@ extension SwiftSDKGenerator {
   func copyTargetSwift(from distributionPath: FilePath, sdkDirPath: FilePath) async throws {
     logGenerationStep("Copying Swift core libraries for the target triple into Swift SDK bundle...")
 
-    for (pathWithinPackage, pathWithinSwiftSDK) in [
-      ("lib/swift", sdkDirPath.appending("usr/lib")),
-      ("lib/swift_static", sdkDirPath.appending("usr/lib")),
-      ("lib/clang", sdkDirPath.appending("usr/lib")),
-      ("include", sdkDirPath.appending("usr")),
+    for (pathWithinPackage, pathWithinSwiftSDK, ignoreIfMissing) in [
+      ("lib/swift", sdkDirPath.appending("usr/lib"), false),
+      ("lib/swift_static", sdkDirPath.appending("usr/lib"), false),
+      ("lib/clang", sdkDirPath.appending("usr/lib"), true),
+      ("include", sdkDirPath.appending("usr"), false),
     ] {
-      try await rsync(from: distributionPath.appending(pathWithinPackage), to: pathWithinSwiftSDK)
+      try await rsync(
+        from: distributionPath.appending(pathWithinPackage),
+        to: pathWithinSwiftSDK, ignoreIfMissing: ignoreIfMissing
+      )
     }
   }
 }
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
@@ -18,7 +18,7 @@ extension SwiftSDKGenerator {
     baseDockerImage: String,
     sdkDirPath: FilePath
   ) async throws {
-    logGenerationStep("Launching a Docker container to copy Swift SDK for the target triple from it...")
+    logger.info("Launching a container to extract the Swift SDK for the target triple...")
     try await withDockerContainer(fromImage: baseDockerImage) { containerID in
       try await inTemporaryDirectory { generator, _ in
         let sdkUsrPath = sdkDirPath.appending("usr")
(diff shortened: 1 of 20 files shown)


MaxDesiatov:
@swift-ci test

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
```diff
@@ -73,6 +73,11 @@ extension SwiftSDKGenerator {
         // But not in all containers, so don't fail if it does not exist.
         if case .ubuntu = targetDistribution {
           subpaths += [("\(targetTriple.archName)-linux-gnu", false)]
+
+          // Custom subpath for armv7
+          if targetTriple.archName == "armv7" {
+            subpaths += [("arm-linux-gnueabihf", false)]
```
MaxDesiatov, line 79: Does Alpine or other Musl-base distributions use this path? I'd assume those support armv7, just thinking how we can future-proof these codepaths to make it easier to support Musl in the future.

xtremekforever: I did a little looking at some Alpine containers and seems like it doesn't follow the Debian scheme of multi-arch by putting stuff at /usr/lib/<arch files>. It looks that it just has the basic ld and libc stuff in /lib:

```bash
# x86_64
/ # ls /lib/*.so*
/lib/ld-musl-x86_64.so.1    /lib/libc.musl-x86_64.so.1

# armhf
~ # ls /lib/*.so*
/lib/ld-musl-armhf.so.1 /lib/libc.musl-armhf.so.1
```

So I guess what would really have to change is the interpreter path to point to something lib `/lib/ld-musl-<arch>.so.1`:

```
extension Triple {
  var interpreterPath: String {
    switch self.archName {
      case "x86_64": return "/lib64/ld-linux-x86-64.so.2"
      case "aarch64": return "/lib/ld-linux-aarch64.so.1"
      case "armv7": return "/lib/ld-linux-armhf.so.3"
      default: fatalError("unsupported architecture \(self.archName)")
    }
  }
}
```

The `arm-linux-gnueabihf` subpath is only for Debian it seems!

MaxDesiatov: Can we make it conditional on Debian somehow and issue a warning or whatever else is suitable for other unrecognized armv7 distributions?

xtremekforever: Do you mean to throw an error if someone tries to create an armv7 Swift SDK when specifying .rhel as the distribution? The block of code you mention above is already under a .ubuntu conditional to only add these special directories for ubuntu distributions.

xtremekforever: Right now if I do something like this:

```bash
$ swift run swift-sdk-generator make-linux-sdk --swift-version 6.0.3-RELEASE --host x86_64-unknown-linux-gnu --target armv7-unknown-linux-gnueabihf --linu
x-distribution-name rhel --with-docker
```

I just get a Docker error, which of course means there is no container available for rhel-ubi9 for armv7:

```bash
Launching a Docker container to copy Swift SDK for the target triple from it...
Error: process exited non-zero: exit(125)
```

So perhaps this situation can be detected up front, and we can throw an error saying that RHEL UBI9 does not support armv7.

xtremekforever:
Right now if I do something like this:

```bash
$ swift run swift-sdk-generator make-linux-sdk --swift-version 6.0.3-RELEASE --host x86_64-unknown-linux-gnu --target armv7-unknown-linux-gnueabihf --linu
x-distribution-name rhel --with-docker
```

I just get a Docker error, which of course means there is no container available for rhel-ubi9 for armv7:

```bash
Launching a Docker container to copy Swift SDK for the target triple from it...
Error: process exited non-zero: exit(125)
```

So perhaps this situation can be detected up front, and we can throw an error saying that RHEL UBI9 does not support armv7. What do you think @MaxDesiatov ?

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Copy.swift
MaxDesiatov: WDYT of logging a warning message at least when non-Ubuntu codepath is triggered for armv7 to point users in the right direction?

MaxDesiatov:
> So perhaps this situation can be detected up front, and we can throw an error saying that RHEL UBI9 does not support armv7. What do you think @MaxDesiatov ?

Yes, that's what I was referring to in the other thread. If we know that generation for other armv7 distributions can't possibly work and we can detect that early, we should log an error message. Or if it could possibly work, but we don't know for sure, then at least a warning message saying this codepath is not fully supported.

xtremekforever:
@MaxDesiatov RHEL and Fedora do not support armv7 and there don't appear to be any plans to support it either:

 - https://fedoraproject.org/wiki/Changes/RetireARMv7
 - https://nullr0ute.com/2022/02/short-history-of-armv7-armhfp-arm32-in-fedora/

So, I've opted to add a new `.distributionDoesNotSupportArchitecture` error to `GeneratorError` that is thrown if you try to create a Swift SDK for a rhel target with armv7 architecture:

`Error: Target Linux distribution RHEL UBI9 does not support the target architecture: armv7`

There really is no way this configuration would work unless someone in a corner of the galaxy is compiling and running RHEL for armv7 for some reason.

MaxDesiatov (approved):
Thanks!

MaxDesiatov:
@swift-ci test

xtremekforever:
@euanh @kateinoigakukun do you guys have a minute to check out this PR this week? The changes are pretty minor I think, and work well as long as you've got build artifacts, so it could be good to have in the generator while I continue working on updating swiftlang/swift to be able to cross-compile for armv7.

kateinoigakukun (approved):

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci build toolchain

Issue #110, opened 2024-06-16
warning: Could not read SDKSettings.json for SDK


toffaletti:
custom SDKs generated with `swift-sdk-generator` cause `swift build` to repeatedly output this warning:

```
warning: Could not read SDKSettings.json for SDK
```

This can be seen in logs from other issues:
https://github.com/apple/swift-sdk-generator/issues/12
https://github.com/apple/swift-sdk-generator/issues/27

xtremekforever:
@MaxDesiatov @toffaletti I did a little research today about this issue and found that this `SDKSettings.json` file is used by the [SwiftCrossCompilers](https://github.com/CSCIX65G/SwiftCrossCompilers) project. Check out the reference file here: https://github.com/CSCIX65G/SwiftCrossCompilers/blob/master/SDKSettings.json

I tried a very minimal file to see what would work to just silence the warning. Including the following content in `SDKSettings.json` in the SDK directory (the ubuntu-jammy.sdk directory, for example) works:

```json
{
  "SupportedTargets": {},
  "Version": "0.0.1",
  "CanonicalName": "linux"
}
```

Is it worth including this file by default in the generated SDKs to silence this warning and make the output cleaner? It seems like this is a feature of swiftpm now and unless it's removed, it's going to search for the file whenever cross-compiling is happening. 

In my SDK I'm adding this empty file to make the compilation output cleaner. It doesn't hurt anything although it's annoying that it has to be there just to make the warning go away.

kateinoigakukun:
The warning is now suppressed in the main branch toolchain https://github.com/swiftlang/swift-driver/pull/1619

PR #185, opened 2025-02-14
Generate SDKSettings.json file for Swift SDKs for Swift 5.9-6.0

xtremekforever:
I know that the warnings about the `SDKSettings.json` file were fixed in Swift 6.1 and later and that issue #110 was closed. However, we are still generating Swift SDKs for 5.9, 5.10, and 6.0 which all give the warnings. 

So, my suggestion is to generate this `SDKSettings.json` file for versions older than 6.1, similar to how the the Linux Static Swift SDKs do it. Here is the file that comes in the `swift-6.0.3-RELEASE_static-linux-0.0.1` Swift SDK for example:

```json
{
  "DisplayName": "Swift SDK for Statically Linked Linux (x86_64)",
  "Version": "0.0.1",
  "VersionMap": {},
  "CanonicalName": "x86_64-swift-linux-musl"
}
```

Taking inspiration from this, I am generating a similar file for Swift SDKs created by the generator:

```json
{
  "CanonicalName" : "x86_64-swift-linux-gnu",
  "DisplayName" : "Swift SDK for Ubuntu Jammy (x86_64)",
  "Version" : "0.0.1",
  "VersionMap" : {

  }
}
```

The ordering is not the same due to how `JSONEncoder()` sorts things, and the formatting is a little different too due to `.prettyPrinted`, but it works just the same. And, the "warning: Could not read SDKSettings.json for SDK at: <path>" warning is successfully suppressed.

This really is a minor change that is more "quality-of-life" then adding any functionality. But, not having those warnings printed when using the generated Swift SDKs are used is a nice to have.

@MaxDesiatov @euanh @kateinoigakukun

### DIFF
--- a/Sources/SwiftSDKGenerator/Extensions/String+hasPrefixIn.swift
+++ b/Sources/SwiftSDKGenerator/Extensions/String+hasPrefixIn.swift
@@ -1,0 +1,10 @@
+extension String {
+    func hasPrefix(in array: [String]) -> Bool {
+        for item in array {
+            if self.hasPrefix(item) {
+                return true
+            }
+        }
+        return false
+    }
+}
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Metadata.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Metadata.swift
@@ -112,4 +112,35 @@ extension SwiftSDKGenerator {
       )
     )
   }
+
+  struct SDKSettings: Codable {
+    var DisplayName: String
+    var Version: String = "0.0.1"
+    var VersionMap: [String: String] = [:]
+    var CanonicalName: String
+  }
+
+  /// Generates an `SDKSettings.json` file that looks like this:
+  /// 
+  /// ```json
+  /// {
+  ///   "CanonicalName" : "<arch>-swift-linux-[gnu|gnueabihf]",
+  ///   "DisplayName" : "Swift SDK for <distribution> (<arch>)",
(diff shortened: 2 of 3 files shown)


Sources/SwiftSDKGenerator/Extensions/String+hasPrefixIn.swift
```diff
@@ -0,0 +1,10 @@
+extension String {
+    func hasPrefix(in array: [String]) -> Bool {
+        for item in array {
+            if self.hasPrefix(item) {
+                return true
+            }
+        }
+        return false
+    }
+}
```
MaxDesiatov, line 10: missing final newline

MaxDesiatov (commented):
Thanks! Mind adding a test for this?

kateinoigakukun (approved):

Sources/SwiftSDKGenerator/Extensions/String+hasPrefixIn.swift
```diff
@@ -0,0 +1,10 @@
+extension String {
+    func hasPrefix(in array: [String]) -> Bool {
```
MaxDesiatov, line 2: This seems a bit more readable with this slightly more verbose name
```suggestion
    func hasAnyPrefix(from array: some String]) -> Bool {
```

MaxDesiatov (approved):
Overall makes sense, just a function naming nit to pick.

MaxDesiatov:
@swift-ci test

euanh (approved):

MaxDesiatov:
@swift-ci test

Issue #116, opened 2024-07-13
Support Debian 12 Bookworm as a distribution to generate Swift SDKs for


xtremekforever:
I have been using the SDK generator to try and build SDKs against Debian 12 Bookworm to be able to cross compile for the Raspberry Pi.

My proposal in this issue is to add support for Debian  12 and even Raspberry Pi OS (if possible) as a distribution that we can download packages and for and build a sysroot with.

Also, if Debian 12 or Raspberry Pi OS 12 is selected, the Swift tool chain for Debian 12 is downloaded if available. Otherwise, the toolchain path must be passed manually.

MaxDesiatov:
Can this be considered done?

xtremekforever:
Yay, thanks guys!

PR #188, opened 2025-02-16
Add support for generating Swift SDKs for Ubuntu 24.04 Noble

xtremekforever:
I was working on adding Debian 12 support for #116, but realized that adding Ubuntu Noble is "low-hanging fruit", since it's very straightforward. The default is still Ubuntu 22.04 Jammy, but this adds the option of generating the Swift SDK for 24.04 Noble now.

```
swift run swift-sdk-generator make-linux-sdk --linux-distribution-version 24.04
```

I have also changed the packages download to get `Packages.xz` instead of `Packages.gz` since it is a smaller file download. I wonder if it is okay to use `xz` here since I noticed that the directories on the Debian mirrors only have `Packages.xz` files available, unlike the Ubuntu mirrors which all have `Packages.gz` AND `Packages.xz` files available.

@MaxDesiatov @euanh

### DIFF
--- a/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
+++ b/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
@@ -202,7 +202,7 @@ extension HTTPClientProtocol {
     \(mirrorURL)/dists/\(ubuntuRelease)\(releaseSuffix)/\(repository)/binary-\(
       targetTriple.arch!
         .debianConventionName
-    )/Packages.gz
+    )/Packages.xz
     """
 
     guard let packages = try await downloadUbuntuPackagesList(
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ The generator also allows cross-compiling between any Linux distributions offici
 | -:             | :-                        | :-                          |
 | macOS (arm64)  | ✅ macOS 13.0+            | ❌                         |
 | macOS (x86_64) | ✅ macOS 13.0+[^1]        | ❌                         |
-| Ubuntu         | ✅ 20.04+                 | ✅ 20.04 / 22.04           |
+| Ubuntu         | ✅ 20.04+                 | ✅ 20.04+                  |
 | RHEL           | ✅ Fedora 39[^2], UBI 9   | ✅ UBI 9                   |
 | Amazon Linux 2 | ✅ Supported              | ✅ Supported[^3]           |
 | Debian 12      | ✅ Supported[^2]          | ✅ Supported[^2][^3]       |
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -200,7 +200,7 @@ extension GeneratorCLI {
     @Option(
       help: """
       Version of the Linux distribution used as a target platform. Available options for Ubuntu: `20.04`, \
-      `22.04` (default when `--linux-distribution-name` is `ubuntu`). Available options for RHEL: `ubi9` (default when \
+      `22.04` (default when `--linux-distribution-name` is `ubuntu`), `24.04`. Available options for RHEL: `ubi9` (default when \
       `--linux-distribution-name` is `rhel`).
       """
     )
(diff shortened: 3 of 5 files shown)


MaxDesiatov:
IIRC problem with xz was that it required additional tooling installed on macOS, and at least documentation updated here in `README.md` to reflect that. Maybe we should also have it listed in a `Brewfile`, especially if we don't have this file already.

MaxDesiatov:
Yes, providing a command to install from
a `Brewfile` in `README.md` seems optimal.

I'd also prefer the generator to fall back to the currently support archive format if xz is not installed.

xtremekforever:
> Yes, providing a command to install from a `Brewfile` in `README.md` seems reasonable.
> 
> I'd also prefer the generator to fall back to the currently support archive format if xz is not installed.

That can be done! And then when we are interacting with the Debian mirrors, I guess we could just throw if xz is not installed and show that it is required for creating a Swift SDK for Debian.

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
```diff
@@ -198,15 +207,16 @@ extension HTTPClientProtocol {
       mirrorURL = ubuntuARM64Mirror
     }
 
+    let xzPath = try await which("xz")
```
MaxDesiatov, line 210: Maybe let's emit a warning or info log message here, letting the user know they can install xz to optimize for download size. @euanh WDYT?

MaxDesiatov (commented):
LGTm

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
xtremekforever: That's reasonable. Let me also add the Brewfile and README content for macOS.

xtremekforever: Here's what I have:

```
2025-02-19T12:33:46-0500 warning org.swift.swift-sdk-generator : [SwiftSDKGenerator] The 'xz' utility was not found in path. Consider installing it for more efficient downloading of package lists.
```

xtremekforever: Huh, running `/usr/bin/which` is just returning an error on macOS for some reason. I might have to change how we do this an manually specify the path for `xz` to `/opt/homebrew/bin`, and just check if the file exists using the file manager instead.

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
```diff
@@ -82,27 +96,39 @@ extension SwiftSDKGenerator {
   ) async throws {
     logger.debug("Parsing Ubuntu packages list...")
 
+    // Find xz path
+    let xzPath = getXzPath()
+    if xzPath == nil {
+      logger.warning("""
+      The 'xz' utility was not found in path. \
```
MaxDesiatov, line 103:
```suggestion
      The `xz` utility was not found in `PATH`. \
```

Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator+Download.swift
```diff
@@ -73,6 +73,20 @@ extension SwiftSDKGenerator {
     ])
   }
 
+  private func getXzPath() -> String? {
+    #if os(macOS)
+      let xzPath = "/opt/homebrew/bin/xz"
```
MaxDesiatov, line 78: This is ARM-specific path and is not the same on Intel. IIRC a portable way to get the path is `brew --prefix xz` invocation or something similar

xtremekforever: Let me try it!!

xtremekforever: Turns out /usr/bin/which was not working because I needed to pass the `environment:` to the ProcessExecutor. When I did that, then the `which()` method worked. I like this better since it is also more portable between macOS and Linux.

MaxDesiatov (approved):

MaxDesiatov:
@swift-ci test

PR #190, opened 2025-02-26
Use `experimental` prefix only for old Swift versions

MaxDesiatov:
`experimental-` prefix is no longer needed with Swift 6.0 and later versions, let's update `README.md` accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,16 +16,23 @@ a newly generated Swift SDK for users to install.
 
 ## Requirements
 
-Usage of Swift SDKs requires Swift 5.9, follow [installation instructions on swift.org](https://www.swift.org/install/) to install it first.
+Usage of Swift SDKs requires Swift 6.0 or later (experimental implementation is available in Swift 5.9/5.10), follow [installation instructions on swift.org](https://www.swift.org/install/) to install it first.
 
-After that, verify that the `experimental-sdk` command is available:
+After that, verify that the `sdk` command is available:
+
+```
+swift sdk list
+```
+
+For Swift 5.9/5.10 add `experimental-` prefix to the command invocation:
+
 
 ```
 swift experimental-sdk list
 ```
 
 The output will either state that no Swift SDKs are available, or produce a list of those you previously had 
-installed, in case you've used the `swift experimental-sdk install` command before.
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
@swift-ci test

xtremekforever:
There is also the message that is printed when the Swift SDK is generated ([here](https://github.com/swiftlang/swift-sdk-generator/blob/9d3a02e9f3b7ce086a96e4cde4c5833391433bba/Sources/SwiftSDKGenerator/Generator/SwiftSDKGenerator%2BEntrypoint.swift#L64C9-L74C10)):

```swift
 print(
          """

          All done! Install the newly generated SDK with this command:
          swift experimental-sdk install \(pathsConfiguration.artifactBundlePath)

          swift build --experimental-swift-sdk \(artifactID)

          """
        )
```

Should this also be changed to use `swift sdk install` and `swift build --swift-sdk` if running on Swift 6.0 or later?

README.md
```diff
@@ -169,7 +182,7 @@ swift run swift-sdk-generator make-linux-sdk --with-docker --from-container-imag
 ## Swift SDK distribution
 
 The `.artifactbundle` directory produced in the previous section can be packaged as a `.tar.gz` archive and redistributed
-in this form. Users of such Swift SDK bundle archive can easily install it with `swift experimental-sdk install`
+in this form. Users of such Swift SDK bundle archive can easily install it with `swift sdk install`
 command, which supports both local file system paths and public `http://` and `https://` URLs as an argument.
```
xtremekforever, line 186: I wonder if there should also be information about generating a checksum for the package. If you try installing a Swift SDK from a remote URL, swiftpm wants you to pass a checksum. For example:

```bash
$ swift sdk install https://download.swift.org/swift-6.0.3-release/static-sdk/swift-6.0.3-RELEASE/sw
ift-6.0.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz
Error: Bundles installed from remote URLs (`https://download.swift.org/swift-6.0.3-release/static-sdk/swift-6.0.3-RELEASE/swift-6.0.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz`) require their checksum passed via `--checksum` option.
The distributor of the bundle must compute it with the `swift package compute-checksum` command and provide it with their Swift SDK installation instructions.
```

xtremekforever (approved):

PR #194, opened 2025-03-02
Fix support for building with `--static-swift-stdlib` for armv7

xtremekforever:
The [atomic library is used in the Swift stdlib](https://github.com/swiftlang/swift/blob/69756635819114a851023aa2e8e6fd87f51d094d/cmake/modules/AddSwift.cmake#L336-L338) for 32-bit architectures. As such, when attempting to compile a binary containing the static Swift stdlib using a Swift SDK for armv7, `-latomic` must be linked. This is not needed for regular cross-compilation, just when passing `--static-swift-stdlib`.

I've added `-latomic` for "armv7" and added a test case in `LinuxRecipeTests` to ensure that it is included in the toolset.json file.

Now, I can build my binary that contains the statically linked Swift stdlib with this command:

```
swift build --swift-sdk 6.0.3-RELEASE_debian_bookworm_armv7 --static-swift-stdlib
```

This generates a rather large binary when also linking other things such as Foundation and Dispatch, as is to be expected:

```
$ pwd
~/tmp/hummingbird-examples/hello
$ du -hs .build/debug/App
123M	.build/debug/App
```

But that's fine, because it works perfectly on the target (Raspberry Pi 2) and provides yet another option for compiling binaries using the Swift SDK.

### DIFF
--- a/Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
+++ b/Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
@@ -129,7 +129,7 @@ public struct LinuxRecipe: SwiftSDKRecipe {
       swiftCompilerOptions.append("-use-ld=lld")
 
       // 32-bit architectures require libatomic
-      if targetTriple.archName == "armv7" {
+      if let arch = targetTriple.arch, arch.is32Bit {
         swiftCompilerOptions.append("-latomic")
       }
 
--- a/Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
+++ b/Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
@@ -128,6 +128,11 @@ public struct LinuxRecipe: SwiftSDKRecipe {
     } else {
       swiftCompilerOptions.append("-use-ld=lld")
 
+      // 32-bit architectures require libatomic
+      if targetTriple.archName == "armv7" {
+        swiftCompilerOptions.append("-latomic")
+      }
+
       if self.hostSwiftSource != .preinstalled {
         toolset.linker = Toolset.ToolProperties(path: "ld.lld")
       }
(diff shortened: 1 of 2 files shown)


Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
```diff
@@ -128,6 +128,11 @@ public struct LinuxRecipe: SwiftSDKRecipe {
     } else {
       swiftCompilerOptions.append("-use-ld=lld")
 
+      // 32-bit architectures require libatomic
+      if targetTriple.archName == "armv7" {
```
MaxDesiatov, line 132: Is this also the case for `i686`? I'm not sure what's the current status of i686 (IIRC there are still some Linux and maybe even BSD distributions that support it?), but curious if it makes sense to keep that path open.

xtremekforever: Yep it also applies for i686, armv5, and armv6 as well. So it could always be checked against any of those architectures. 

I was thinking in the future if the generator is upgraded to use an external sysroot as an option, then it can support pretty much any arch that is supported in Swift as well.

So, do you want to add more arches to this check here?

euanh (approved):

Sources/SwiftSDKGenerator/SwiftSDKRecipes/LinuxRecipe.swift
MaxDesiatov: WDYT about adding `var is32Bit: Bool` to `Triple` and checking for that here?

MaxDesiatov: Ah, I see it's vendored from `swift-driver`. But good news is that it's already there IIUC: https://github.com/swiftlang/swift-driver/blob/5cc1fb1a51f5ca3c4be88be0a9efd263eab83d57/Sources/SwiftDriver/Utilities/Triple.swift#L815

Issue #189, opened 2025-02-25
Unable to create Bookworm SDK


dscreve:
Just enter the following command : 

$> swift run swift-sdk-generator make-linux-sdk --with-docker --from-container-image swift:6.0.3-bookworm:latest --sdk-name 6.0.3-bookworm

Result : 
warning: failed to retrieve search paths with pkg-config; maybe pkg-config is not installed
warning: couldn't find pc file for sqlite3
Building for debugging...
[1/1] Write swift-version--58304C5D6DBC2206.txt
Build of product 'swift-sdk-generator' complete! (2.18s)
2025-02-25T19:48:51+0100 info org.swift.swift-sdk-generator : [SwiftSDKGenerator] Downloading required toolchain packages...
2025-02-25T19:48:51+0100 info org.swift.swift-sdk-generator : [SwiftSDKGenerator] Using downloaded artifacts from cache
2025-02-25T19:48:51+0100 info org.swift.swift-sdk-generator : [SwiftSDKGenerator] Launching a container to extract the Swift SDK for the target triple...
Error: process exited non-zero: exit(125)

xtremekforever:
Looks like you have a typo in your tag name. It should be:

`--from-container-image swift:6.0.3-bookworm`

Without the last ":latest" part.

MaxDesiatov:
Hi @dscreve, would you be able to confirm that the issue is resolved for you when using the proposed tag name?

dscreve:
Hello,

This works fine, but If I use an amd64 image on Mac M1, it does not works. Only arm64 image works?

David

dscreve:
Thanks ! this works fine.

What are the accepted values for --parameters ? readme.md documentation seems to be outdated and incomplete.

David

xtremekforever:
Um....I think that CryptoKit is not available on Linux platforms...am I correct @MaxDesiatov ?

MaxDesiatov:
You should use [swift-crypto](https://github.com/apple/swift-crypto/) instead, which will switch between CryptoKit and the OSS shims over BoringSSL, depending on the OS you're targeting.

euanh:
You can ignore these warnings:

> warning: Could not read SDKSettings.json for SDK at: /Users/dscreve/Library/org.swift.swiftpm/swift-sdks/6.0.3-bookworm-amd64.artifactbundle/6.0.3-bookworm-amd64/x86_64-unknown-linux-gnu/ubuntu-jammy.sdk

[...]

PR #196, opened 2025-03-07
Enable host toolchain on macOS by default, improve `README.md`

xtremekforever:
I put this against #189 since it adds some documentation to help with common issues people run into. But primarily, this PR sets `--host-toolchain` on macOS by default, since most people are using Xcode and not the Swift OSS toolchain to run the generator and build using generated Swift SDKs.


MaxDesiatov (approved):
Makes sense to me. I'm not fond of requiring `--target` with a full triple name these days,  as it's mostly CPU arch that users care about, and other triple components can (and should be) inferred. But that's something to do in the future, not blocking for this PR.

MaxDesiatov:
@swift-ci test

xtremekforever:
Also, thought I would share. I was talking to the author of SwiftToolkit.dev yesterday and he ran into the same issue as others had with using Swift SDKs generated on macOS without the host toolchain. I explained to him that `--host-toolchain` was needed if he was not using the Swift OSS toolchain. This is what inspired this change, since most people are running the Xcode Swift toolchain.

However, he did write a nice post about running Swift on the Raspberry Pi, which is well supported now for aarch64 PIs:

https://swifttoolkit.dev/posts/r-pi

xtremekforever:
> Makes sense to me. I'm not fond of requiring `--target` with a full triple name these days, as it's mostly CPU arch that users care about, and other triple components can (and should be) inferred. But that's something to do in the future, not blocking for this PR.

Yeah, especially since we only ever generate Swift SDKs to cross-compile to Linux it doesn't seem useful to have to include the whole triple all the time. So something like this could be awesome in the future:

```bash
--target x86_64
--target aarch64
--target armv7
```

Or, even let it take alternate names too:

```bash
--target amd64
--target arm64
--target arm    # could default to armv7 since this is the latest arm "32-bit" supported
```

MaxDesiatov:
We previously had this called `--arch` IIRC, not sure when exactly it morphed into a full triple option named `--target`. @kateinoigakukun maybe it was WASI related?

xtremekforever:
> We previously had this called `--arch` IIRC, not sure when exactly it morphed into a full triple option named `--target`. @kateinoigakukun maybe it was WASI related?

Was it `--target-arch`? 

```
  --target-arch <target-arch>
                          Deprecated. Use `--target` instead (values: arm, armeb, aarch64, aarch64e, aarch64_be,
                          aarch64_32, arc, avr, bpfel, bpfeb, hexagon, m68k, mips, mipsel, mips64, mips64el, msp430, ppc,
                          ppc64, ppc64le, r600, amdgcn, riscv32, riscv64, sparc, sparcv9, sparcel, systemz, tce, tcele,
                          thumb, thumbeb, i386, x86_64, xcore, nvptx, nvptx64, le32, le64, amdil, amdil64, hsail,
                          hsail64, spir, spir64, kalimba, shave, lanai, wasm32, wasm64, renderscript32, renderscript64,
                          xtensa)
```

kateinoigakukun:
It makes sense to me to alias known arch names to recipe-specific defaults. I made the option to accept a full triple to be able to handle "environment" part difference but I understand it's not always necessary.

xtremekforever:
> It makes sense to me to alias known arch names to recipe-specific defaults. I made the option to accept a full triple to be able to handle "environment" part difference but I understand it's not always necessary.

So, I wonder if maybe using `--target-arch` would not be deprecated but instead an alternative for being able to pass just the arch, so you could tell it `aarch64` or `x86_64`. The problem is that it'll have to translate them to the `*-unknown-linux-gnu` but maybe it's doable. What do you think @MaxDesiatov ? 

Then if you need something that's not a Linux platform, just use the whole triple argument instead...

MaxDesiatov:
> The problem is that it'll have to translate them to the `*-unknown-linux-gnu` but maybe it's doable.

Surely doable, ideally accounting for possible future Musl support (hence `-gnu` is not the only available environment/ABI). Not required for this PR of course.

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

xtremekforever:
Right now if I give the swift-sdk-generator the `--target-arch` param it'll tell me to use `--target` instead:

```
2025-03-11T14:20:19-0400 warning org.swift.swift-sdk-generator : [GeneratorCLI] deprecated: Please use `--target aarch64-unknown-linux-gnu` instead of `--target-arch aarch64`
```

And the end result is a Swift SDK for the wrong arch and not what I wanted:

```
swift build --experimental-swift-sdk 6.0.3-RELEASE_ubuntu_jammy_x86_64
```

So my question is- should we revive the use of the --target-arch param as an alternative to the use of --target, and of course it'll just default to *-linux-gnu as it does now?

```swift
    func deriveTargetTriple(hostTriple: Triple) -> Triple {
      if let target = generatorOptions.target {
        return target
      }
      if let arch = generatorOptions.targetArch {
        let target = Triple(arch: arch, vendor: nil, os: .linux, environment: .gnu)
        appLogger.warning(
          "deprecated: Please use `--target \(target.triple)` instead of `--target-arch \(arch)`")
      }
      return Triple(arch: hostTriple.arch!, vendor: nil, os: .linux, environment: .gnu)
    }
```

README.md
```diff
@@ -116,6 +116,43 @@ for Ubuntu Jammy and Swift 5.9 this would be `swift:5.9-jammy-slim`. If you'd li
...
+swift run swift-sdk-generator make-linux-sdk --target-arch x86_64
+```
+
+This will default to building the Swift SDK for `x86_64-unknown-linux-gnu`. To build for other
+platforms and environments, supply the `--target` flag with the full target triple instead.
+
+The Linux distribution name and version can also be passed to change from the default of Ubuntu Jammy:
+
+```bash
+swift run swift-sdk-generator make-linux-sdk --linux-distribution-name ubuntu --linux-distribution-version noble
```
MaxDesiatov, line 135: Sorry for pestering with more requests, but writing up new documentation is a good opportunity to look at user experience from a different angle. What's your opinion on removing the `linux-` prefix from these option names? Since they're already on the `make-linux-sdk` subcommand, it's already obvious what they're referring to. IIUC the prefixes are vestigial from the ancient times when we didn't have separate subcommands.

```suggestion
swift run swift-sdk-generator make-linux-sdk --distribution-name ubuntu --distribution-version noble
```

README.md
```diff
@@ -116,6 +116,43 @@ for Ubuntu Jammy and Swift 5.9 this would be `swift:5.9-jammy-slim`. If you'd li
...
+### Host Toolchain
+
+The host toolchain is not included in the generated Swift SDK by default on Linux to match the behavior
+of the [Static Linux Swift SDKs](https://www.swift.org/documentation/articles/static-linux-getting-started.html)
+downloadable from [swift.org](https://www.swift.org/install/). However, on macOS, since most users are using Xcode
+and are likely not using the Swift OSS toolchain to build and run Swift projects, the Swift host toolchain
+is included by *default*. This default behavior can be changed by passing  `--no-host-toolchain`:
+
+```bash
+swift run swift-sdk-generator make-linux-sdk --no-host-toolchain --target x86_64
```
MaxDesiatov, line 147: Should this be `--target-arch`?
```suggestion
swift run swift-sdk-generator make-linux-sdk --no-host-toolchain --target-arch x86_64
```

README.md
```diff
@@ -116,6 +116,43 @@ for Ubuntu Jammy and Swift 5.9 this would be `swift:5.9-jammy-slim`. If you'd li
...
+```
+
+Or, if on Linux, and desiring to generate the Swift SDK with the host toolchain included, add `--host-toolchain`:
+
+```bash
+swift run swift-sdk-generator make-linux-sdk --host-toolchain --target aarch64
```
MaxDesiatov, line 153: Ditto?

```suggestion
swift run swift-sdk-generator make-linux-sdk --host-toolchain --target-arch aarch64
```

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -148,17 +148,30 @@ extension GeneratorCLI {
     var host: Triple? = nil
 
     @Option(
-      help: """
-        The target triple of the bundle. The default depends on a recipe used for SDK generation. Pass `--help` to a specific recipe subcommand for more details.
-        """
-    )
+      help:
+        "The target triple of the bundle. The default depends on a recipe used for SDK generation.")
```
MaxDesiatov, line 152:
```suggestion
        "The target triple of the bundle. The default depends on a recipe used for SDK generation."
      )
```

README.md
xtremekforever: Ah yes, I missed that!! DX

README.md
xtremekforever: I think it's a great idea! We should go ahead and do it!

MaxDesiatov: @euanh do you know of any scripts/plugins that could rely on option names with prefixes? Do you think we need to formally deprecate them?

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -148,17 +148,31 @@ extension GeneratorCLI {
     var host: Triple? = nil
 
     @Option(
-      help: """
-        The target triple of the bundle. The default depends on a recipe used for SDK generation. Pass `--help` to a specific recipe subcommand for more details.
-        """
-    )
+      help:
+        "The target triple of the bundle. The default depends on a recipe used for SDK generation."
+      )
```
MaxDesiatov, line 153:
```suggestion
    )
```

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

xtremekforever:
@MaxDesiatov what's the plan here- do you need any more changes or reviews for this? I'm hoping to be able to include these changes if they are acceptable? :)

Sources/GeneratorCLI/GeneratorCLI.swift
```diff
@@ -149,17 +149,31 @@ extension GeneratorCLI {
...
     var target: Triple? = nil
 
     @Option(help: "Deprecated. Use `--host` instead")
     var hostArch: Triple.Arch? = nil
-    @Option(help: "Deprecated. Use `--target` instead")
+    @Option(
+      help: """
+        The target arch of the bundle. The default depends on a recipe used for SDK generation.
+        If this is passed, the target triple will default to `<target-arch>-unknown-linux-gnu`.
+        Use the `--target` param to pass the full target triple if needed.
```
MaxDesiatov, line 163:
```suggestion
        The target arch of the bundle. The default depends on a recipe used for SDK generation. \
        If this is passed, the target triple will default to `<target-arch>-unknown-linux-gnu`. \
        Use the `--target` param to pass the full target triple if needed.
```

MaxDesiatov:
@swift-ci test

PR #204, opened 2025-04-03
Update defaults for Swift 6.1-RELEASE

xtremekforever:
- Defaulted `swiftVersion` to "6.1-RELEASE".
 - Added EndToEndTests for Ubuntu and RHEL for Swift61, including some additional tests for Ubuntu Jammy.
 - Fixed missing /lib and /lib64 directories in Ubuntu Noble when building from deb packages. If these directories don't exist, they are symlinked to the correct directories in /usr, which is consistent to what happens when copying from the Docker containers.
 - Added the fix for downloading Swift for the target that adds the platform name to the cache file, to avoid corruption of the artifacts: 
     * image

 ALSO: I did not include EndToEndTests for Fedora39 for Swift 6.1 since the container is non-existent, likely because it's EOL.

### DIFF
--- a/Sources/GeneratorCLI/GeneratorCLI.swift
+++ b/Sources/GeneratorCLI/GeneratorCLI.swift
@@ -113,7 +113,7 @@ extension GeneratorCLI {
     var swiftBranch: String? = nil
 
     @Option(help: "Version of Swift to supply in the bundle.")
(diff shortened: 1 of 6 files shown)


MaxDesiatov:
@swift-ci test

PR #207, opened 2025-05-14
Bump `// swift-tools-version` to 5.9

MaxDesiatov:
All of the CI jobs are using Swift 5.9 or later at this point.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.8
+// swift-tools-version: 5.9
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription


MaxDesiatov:
@swift-ci test

euanh:
Has the WASM CI moved away from 5.8?   I remember @kateinoigakukun having to revert some changes which depended on post-5.8 features in the past.

euanh:
Are there any backwards-compatibility workarounds which we can remove when we no longer have to support 5.8?

kateinoigakukun:
Wasm CI jobs on ci.swift.org already upgraded the bootstrap toolchain to 5.9, so it's ok.

MaxDesiatov:
> Are there any backwards-compatibility workarounds which we can remove when we no longer have to support 5.8?

The only one, off the top of my head, is that we can start using `package` instead of `public` 🙂

PR #208, opened 2025-05-14
Generate Embedded Swift SDK for WASI

MaxDesiatov:
This allows targeting WASI with Embedded Swift, enabling use of a subset of Swift Concurrency in this mode, and also hiding most if not all of unsafe flags in toolsets. This way, users of Embedded Swift can build their packages with a similar `swift build --swift-sdk` command as the rest of users of WASI.

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -11,7 +11,7 @@
   "lineBreakBeforeControlFlowKeywords" : false,
   "lineBreakBeforeEachArgument" : true,
   "lineBreakBeforeEachGenericRequirement" : false,
-  "lineLength" : 100,
+  "lineLength" : 120,
   "maximumBlankLines" : 1,
   "multiElementCollectionTrailingCommas" : true,
   "noAssignmentInExpressions" : {
--- a/Package.resolved
+++ b/Package.resolved
@@ -5,8 +5,17 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/swift-server/async-http-client.git",
       "state" : {
-        "revision" : "16f7e62c08c6969899ce6cc277041e868364e5cf",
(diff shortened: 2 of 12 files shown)


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

MaxDesiatov:
@euanh this might be interesting for you as we potentially could do the same with for Linux. IDK how well Embedded Swift works with Glibc, but certainly should be possible with Musl.

euanh:
> @euanh this might be interesting for you as we potentially could do the same with for Linux. IDK how well Embedded Swift works with Glibc, but certainly should be possible with Musl.

That's an interesting idea!   Definitely worth trying it.    It looks like the main functional change here is adding extra compiler flags.  Is that right?

MaxDesiatov:
> > @euanh this might be interesting for you as we potentially could do the same with for Linux. IDK how well Embedded Swift works with Glibc, but certainly should be possible with Musl.
> 
> That's an interesting idea! Definitely worth trying it. It looks like the main functional change here is adding extra compiler flags. Is that right?

Yes, also adding one more Swift SDK with its own metadata and its own toolset within the same bundle. As a lot of directories overlap between embedded and non-embedded and can be reused, we rely on this recent SwiftPM change to enable that https://github.com/swiftlang/swift-package-manager/pull/8638

euanh:
https://github.com/swiftlang/swift-package-manager/pull/8638 is a nice improvement.

kateinoigakukun (approved):

PR #209, opened 2025-05-14
Bump dependencies in `Package.resolved`

MaxDesiatov:
At the same time, `Package.swift` is aligned more closely to `update-checkout-config.json` from the toolchain repository


MaxDesiatov:
@swift-ci test

PR #212, opened 2025-05-16
Fix use of absolute paths to Swift SDK metatadata

MaxDesiatov:
The generator incorrectly recorded absolute paths to Swift SDK metadata in bundle archive metadata, which is fragile and also can't be handled by SwiftPM.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

PR #227, opened 2025-07-15
Enable `amazonlinux2` in `pull_request.yml`

MaxDesiatov:
Due to frequent regressions with Amazon Linux 2 we should test it on CI.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   tests:
     name: Test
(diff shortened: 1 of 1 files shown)


jakepetroules:
This isn't possible, GitHub Actions doesn't support Amazon Linux 2.

MaxDesiatov:
@swift-ci test

jakepetroules:
> This isn't possible, GitHub Actions doesn't support Amazon Linux 2.

With swiftlang/github-workflows#151, this should work now.

PR #229, opened 2025-07-18
[WebAssembly] Create `clang/lib/wasip1` symlink only if it doesn't exist

kateinoigakukun:
We are renaming `wasm32-unknown-wasi` to `wasm32-unknown-wasip1` for non-Embedded targets, so we will no longer have the `clang/lib/wasi`.


MaxDesiatov:
@swift-ci test

PR #232, opened 2025-08-10
Add support for generating Swift SDKs with multiple host triples

jakepetroules:
This is necessary for correctness on macOS, where the concept of multi-arch binaries exists, and all swift.org toolchains for macOS are multi-arch.

### DIFF
--- a/Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
+++ b/Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
@@ -18,6 +18,8 @@ import struct SystemPackage.FilePath
 extension Triple {
   fileprivate var llvmBinaryURLSuffix: String {
     switch (self.os, self.arch) {
+    case (.win32, .aarch64): return "aarch64-pc-windows-msvc"
+    case (.win32, .x86_64): return "x86_64-pc-windows-msvc"
     case (.linux, .aarch64): return "aarch64-linux-gnu"
     case (.linux, .x86_64): return "x86_64-linux-gnu-ubuntu-22.04"
     case (.macosx, .aarch64): return "arm64-apple-darwin22.0"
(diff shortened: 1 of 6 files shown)


Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
```diff
@@ -121,3 +133,10 @@ struct DownloadableArtifacts: Sendable {
     )
   }
 }
+
+struct StringError: Error, CustomStringConvertible {
```
MaxDesiatov, line 137: I don't see any value in stringly typed APIs in Swift, especially for errors. These are harder to test against, and go against the grain of the static typing approach that Swift takes. IMO a nested `enum Error` within `DownloadableArtifacts` with explicit cases for each error path would work, and would make it easier to adopt typed throws in the future.

jakepetroules:
@swift-ci test

Tests/SwiftSDKGeneratorTests/SwiftSDKRecipes/LinuxRecipeTests.swift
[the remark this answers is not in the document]
jakepetroules: Done

Sources/SwiftSDKGenerator/Artifacts/DownloadableArtifacts.swift
jakepetroules: Done, added an error enum.

Tests/SwiftSDKGeneratorTests/SwiftSDKRecipes/LinuxRecipeTests.swift
```diff
@@ -154,8 +154,37 @@ final class LinuxRecipeTests: XCTestCase {
...
+  }
+
+  func testItemsToDownloadForMacOSHost_multiArch() throws {
+    try testItemsToDownloadForMacOSHost(hostTriples: [Triple("arm64-apple-macos"), Triple("x86_64-apple-macos")])
+  }
+
+  private func testItemsToDownloadForMacOSHost(
+    hostTriples: [Triple],
+    validate: ((Result<Void, Error>) throws -> Void)? = nil
+  ) throws {
```
MaxDesiatov, line 187: It took me a minute to track down logic for `validate`, would you mind adding a doc comment here?

```suggestion
  /// ...
  /// - Parameters:
  ///   - hostTriples: ...
  ///   - validate: ...
  /// - Throws: ...
  private func testItemsToDownloadForMacOSHost(
    hostTriples: [Triple],
    validate: ((Result<Void, Error>) throws -> Void)? = nil
  ) throws {
```

MaxDesiatov (approved):
Thanks! LGTM modulo nits for line length and a clarifying doc comment

jakepetroules:
@swift-ci test

Tests/SwiftSDKGeneratorTests/SwiftSDKRecipes/LinuxRecipeTests.swift
jakepetroules: Done

jakepetroules:
@swift-ci test

PR #242, opened 2025-09-04
Make targetSwiftPackagePath optional for WASM SDK generation

jakepetroules:
This allows generating WASM SDKs that do not have support for targeting Swift.


jakepetroules:
@swift-ci test macOS

PR #244, opened 2025-09-09
WebAssemblyRecipe:: Enable embedded target for wasip1-threads as well

kateinoigakukun:
There is no reason to limit Embedded Swift to only the non-threads variant. The limitation made some tests in JavaScriptKit a bit more complicated, see e.g. https://github.com/swiftwasm/JavaScriptKit/pull/443/files#diff-2c1174f0c456f8536c09dce2f943c43287a6a207ef08054d6aeddf5f00efa040R86


MaxDesiatov:
@swift-ci test

kateinoigakukun:
build-script still doesn't build swift-sdk for the target but swiftwasm/swiftwasm-build does.

PR #251, opened 2025-12-06
Support passing distribution version names for Ubuntu/Debian, add unit tests

### LINKED ISSUE #235: Allow passing distribution version name instead of version number (i.e., noble instead of 24.04)
One thing I've noticed with using the generator is that you must provide the `--distribution-version` as the version number, and cannot use the code name as an alternative:

```console
$ swift run swift-sdk-generator make-linux-sdk --distribution-version noble
[1/1] Planning build
Building for debugging...
[1/1] Write swift-version--149EBBC1E6AAC1E9.txt
Build of product 'swift-sdk-generator' complete! (13.26s)
Error: Linux distribution `ubuntu` with version `noble` is not supported by this generator.
```

However, Swift SDKs are output using the code names by default:

```console
$ swift sdk list
6.1-RELEASE_ubuntu_jammy_aarch64
6.1.2-RELEASE_ubuntu_jammy_aarch64
```

This causes a problem for scripts, since now you need to have the distribution version number as well as the codename in the script:

```bash
DISTRO_NAME=ubuntu
DISTRO_VERSION=jammy
DISTRO_VERSION_GENERATOR=22.04
```

Although it is possible to pass the SDK name when running the generator, it seems like it would still be nicer to have this as an option to be able to pass the version code name to the generator.

xtremekforever:
This fixes #235. With this change you can now pass the code name of the version to the generator:

`swift run swift-sdk-generator make-linux-sdk --distribution-name ubuntu --distribution-version noble`
`swift run swift-sdk-generator make-linux-sdk --distribution-name Debian --distribution-version trixie`

Also, as a part of attempting to improve test coverage of the generator, I've added `LinuxDistributionTests` built on Swift Testing to complete coverage of `LinuxDistribution`. It is only compiled/run on Swift 6.0 and later, so it should always be invoked on the current CIs. Once swift-tools-version is set to 6.0 the `#if canImport(Testing)` wrapper can be removed!

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   tests:
     name: Test
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.2
     with:
       linux_os_versions: '["amazonlinux2", "bookworm", "jammy", "focal", "noble", "rhel-ubi9"]'
       linux_swift_versions: '["6.0", "6.1", "nightly-6.2", "nightly-main"]'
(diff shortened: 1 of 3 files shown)


MaxDesiatov (approved):
Thanks!

PR #273, opened 2026-09-19
Default to Swift 6.4.0-RELEASE in generator, change default Ubuntu to 24.04

xtremekforever:
Pretty simple change. I guess it would be good to spend a little time on the EndToEndTests- I see that we don't have coverage for Swift 6.3 yet, but having coverage for 6.3 and 6.4.0 would be good.


MaxDesiatov (approved):
Thank you!

MaxDesiatov:
@swift-ci test

